PR #58, opened 2026-04-04
fix: resolve pytest collection crash in test_data_sources.py

### LINKED ISSUE #54: [BUG] Pytest crashes during collection due to test_data_sources.py
Running `pytest` on a fresh clone crashes the collection phase entirely because of hardcoded logic in `test_data_sources.py`. The file statically injects a local hardcoded path (`/home/shashank/Desktop/sktime-mcp/src`) and catches `ImportError` exceptions by throwing a `sys.exit(1)`, which violently terminates the entire `pytest` process.

**To Reproduce**
1. Clone the repository into a fresh environment
2. Run `pytest`
3. Notice that the session terminates with a system exit before any tests actually run.

**Expected behavior**
Test collection should succeed and tests should run normally. `sys.path` should be resolved dynamically (or rely on `pip install -e .`), and `sys.exit(1)` should be replaced with a standard Python `raise` command so `pytest` can catch the exception gracefully without dying.

*(I have this fixed locally and can open a quick PR for it!)*

Sumit-ai-dev:
## Description
Running `pytest` on a fresh clone crashes the collection phase entirely due to hardcoded logic in `tests/test_data_sources.py`.

**Root cause:**
- `sys.path.insert(0, '/home/shashank/Desktop/sktime-mcp/src')`  hardcoded absolute path to a local machine
- `sys.exit(1)` inside an `except ImportError` block  violently terminates the entire `pytest` process during collection, not just the one test

**Changes:**
- Replaced the hardcoded `sys.path` with a dynamic `os.path` resolution relative to the test file location
- Replaced `sys.exit(1)` with a standard `raise` so `pytest` handles failures gracefully

**Verified:** Full `pytest tests/` now runs cleanly  39 passed, 0 crashes.

**Fixes #54**

### DIFF
--- a/src/sktime_mcp/registry/interface.py
+++ b/src/sktime_mcp/registry/interface.py
@@ -47,9 +47,9 @@ class EstimatorNode:
             "module": self.module,
             "tags": self.tags,
             "hyperparameters": self.hyperparameters,
-            "docstring": self.docstring[:500]
-            if self.docstring
-            else None,  # L-1: Truncate docstring to 500 characters, we can also try summarization
+            "docstring": (
+                self.docstring[:500] if self.docstring else None
+            ),  # L-1: Truncate docstring to 500 characters, we can also try summarization
         }
 
     def to_summary(self) -> dict[str, Any]:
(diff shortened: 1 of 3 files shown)


Shashankss1205:
Please pull the latest commit and then make this PR, also try to ensure the ruff tests pass.
Thanks!

Sumit-ai-dev:
@Shashankss1205 Ready for review. :D

Shashankss1205:
already solved in your another PR

PR #71, opened 2026-04-04
fix: preserve database dialect in SQL connection string sanitization

### LINKED ISSUE #70: [BUG] SQLAdapter _sanitize_connection_string drops database dialect from metadata
### Describe the bug

In `src/sktime_mcp/data/adapters/sql_adapter.py`, the `_sanitize_connection_string` method is responsible for masking credentials before storing the connection info in the data source metadata.

However, the current implementation uses a destructive `split("@")` that blindly replaces everything before the `@` with `***`. This accidentally destroys the database dialect/protocol scheme (e.g., `postgresql://`).

**Buggy code (lines 168-175):**
```python
def _sanitize_connection_string(self, conn_string: str) -> str:
    if "@" in conn_string:
        parts = conn_string.split("@")
        if len(parts) == 2:
            return f"***@{parts[1]}"
    return conn_string
```

**Example:**
- Input: `postgresql://user:pass@host:5432/db`
- - Current output: `***@host:5432/db` (INCORRECT)
- - - Expected output: `postgresql://***@host:5432/db` (CORRECT)
### Why this matters for LLM integration

The metadata returned by `load_data_source` is a primary information channel for the MCP LLM agent. When the dialect string (`postgresql://`) is stripped, the LLM agent loses the only machine-readable hint about which SQL flavour (PostgreSQL, MySQL, SQLite, etc.) it is connected to. This can cause the agent to generate incorrect SQL dialect syntax.

### Steps to reproduce

```python
from sktime_mcp.data.adapters.sql_adapter import SQLAdapter
adapter = SQLAdapter({})
print(adapter._sanitize_connection_string("postgresql://user:pass@host:5432/db"))
# Prints: ***@host:5432/db  (dialect is lost!)
```

### Expected behavior

The sanitization should mask only the `username:password` portion and preserve the protocol:
- `postgresql://user:pass@host:5432/db` -> `postgresql://***@host:5432/db`
- - `mysql://admin:secret@192.168.1.1/prod` -> `mysql://***@192.168.1.1/prod`
- - - `sqlite:///local.db` -> `sqlite:///local.db` (no credentials, unchanged)
### Proposed fix

Split on `://` first to extract the protocol, then mask only the credentials:

```python
def _sanitize_connection_string(self, conn_string: str) -> str:
    if "@" in conn_string:
        try:
            protocol_auth, rest = conn_string.split("@", 1)
            if "://" in protocol_auth:
                protocol, _ = protocol_auth.split("://", 1)
                return f"{protocol}://***@{rest}"
            return f"***@{rest}"
        except Exception:
            return f"***@{conn_string.split('@')[-1]}"
    return conn_string
```

I am happy to open a PR with this fix!

Sumit-ai-dev:
### Description

This PR fixes an issue where `_sanitize_connection_string` would aggressively split on `@` and replace everything before it with `***`, destroying the database dialect (e.g., `postgresql://`). This caused the LLM agent to lose context on what SQL dialect to use when reading metadata.

The new logic preserves the protocol scheme before applying the mask.

Fixes #70

### DIFF
--- a/src/sktime_mcp/data/adapters/sql_adapter.py
+++ b/src/sktime_mcp/data/adapters/sql_adapter.py
@@ -167,11 +167,16 @@ class SQLAdapter(DataSourceAdapter):
 
     def _sanitize_connection_string(self, conn_string: str) -> str:
         """Remove credentials from connection string for metadata."""
-        # Hide password in connection string
+        # Hide password in connection string but preserve the dialect/protocol
         if "@" in conn_string:
-            parts = conn_string.split("@")
-            if len(parts) == 2:
-                return f"***@{parts[1]}"
+            try:
+                protocol_auth, rest = conn_string.split("@", 1)
+                if "://" in protocol_auth:
+                    protocol, _ = protocol_auth.split("://", 1)
+                    return f"{protocol}://***@{rest}"
+                return f"***@{rest}"
+            except Exception:
+                return f"***@{conn_string.split('@')[-1]}"
         return conn_string
 
     def validate(self, data: pd.DataFrame) -> tuple[bool, dict[str, Any]]:


Sumit-ai-dev:
@Shashankss1205  ready to review :)

PR #76, opened 2026-04-04
Fix #68: Make export_code tool dataset-aware

### LINKED ISSUE #68: [ENH] export_code tool hardcodes airline dataset in fit/predict example
While using the `export_code` tool with `include_fit_example=True`, I noticed that the generated example code always uses `load_airline` regardless of which dataset the estimator was actually used with. This can be confusing when working with other datasets like sunspots or custom data loaded via `load_data_source`.

The `export_code_tool` in `src/sktime_mcp/tools/codegen.py` should generate the example code based on the actual dataset that was used. We could:
- Accept an optional `dataset` parameter in `export_code_tool` and use it to generate the correct `from sktime.datasets import load_<dataset>` line.
- Fall back to `load_airline` if no dataset info is available, to keep backward compatibility.

One option is to store the dataset name in the handle metadata when `fit_predict` is called, and then read it back during code generation. But that adds coupling between tools, so passing it as an optional parameter feels cleaner and simpler for now.

**Additional context**
The hardcoded section is around line 256-268 in `codegen.py`. This is a small change but makes the generated code much more useful and accurate.

Abhishek9639:
#### Reference Issues/PRs

Fixes #68

The `export_code` tool was hardcoding `load_airline` in the generated fit/predict example regardless of which dataset was actually used. This could be confusing when working with other datasets like sunspots or shampoo.

This PR adds an optional `dataset` parameter to `export_code_tool`. When provided, it looks up the correct loader function from `DEMO_DATASETS` and generates the matching import (e.g. `from sktime.datasets import load_sunspot`). If no dataset is passed, it falls back to `load_airline` to keep backward compatibility.

Changes:
- `codegen.py`: Added `dataset` param and dynamic dataset resolution logic.
- `server.py`: Added `dataset` to the tool schema and forwarded it in the dispatch.

No

- The dataset resolution logic in `codegen.py` (lines 244-253) - making sure the fallback behavior is clean
- The schema addition in `server.py` - making sure the description is clear enough for LLM tool use

#### Any other comments?

Small focused change, only 2 files modified. All existing tests pass (41/42, the 1 failure is a pre-existing async test unrelated to this PR).

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/CONTRIBUTORS.md)

##### For new estimators

### DIFF
--- a/src/sktime_mcp/server.py
+++ b/src/sktime_mcp/server.py
@@ -309,6 +309,14 @@ async def list_tools() -> list[Tool]:
                         "description": "Whether to include a fit/predict example (default: false)",
                         "default": False,
                     },
+                    "dataset": {
+                        "type": "string",
+                        "description": (
+                            "Optional dataset name for the fit example "
+                            "(e.g. 'airline', 'sunspots'). "
+                            "Defaults to 'airline' if omitted."
+                        ),
+                    },
                 },
                 "required": ["handle"],
             },
@@ -614,6 +622,7 @@ async def call_tool(name: str, arguments: dict[str, Any]) -> list[TextContent]:
                 arguments["handle"],
                 arguments.get("var_name", "model"),
                 arguments.get("include_fit_example", False),
+                arguments.get("dataset"),
             )
         elif name == "load_data_source":
             result = load_data_source_tool(arguments["config"])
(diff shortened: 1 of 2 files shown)


Abhishek9639:
Greetings @Shashankss1205,
Ready for review whenever you get a chance. 
Thanks

Shashankss1205 (approved):

PR #82, opened 2026-04-05
[BUG] Fix thread-unsafe singleton initialization in factory functions

direkkakkar319-ops:

All five singleton factory functions (get_registry, get_handle_manager, get_executor, get_tag_resolver, get_composition_validator) used a simple if None check with no locking, making them unsafe under concurrent access. When fit_predict_async offloads work to thread pool executors via loop.run_in_executor(), multiple threads can race through the if _instance is None check simultaneously, creating duplicate instances and leading to inconsistent state (e.g., duplicate registry loads, split handle stores).
This PR applies the double-checked locking pattern to all five singletons using threading.Lock
```python
if _instance is None:          # Fast path
    with _lock:                # Serialize only during first initialization
        if _instance is None:  # Re-check under lock
            _instance = Class()
```

* Each file receives two minimal changes:
1. `import threading` added to `imports`
2. A module-level `threading.Lock()` and the double-checked locking pattern in the `factory function`

No. threading is part of the Python standard library.

1. Correctness of the double-checked locking pattern across all five files
2. Whether any of these singletons should additionally protect their internal mutable state with locks for thread-safe mutation, beyond just safe initialization   .

#### Any other comments?
Double-checked locking was used instead of `functools.lru_cache(maxsize=1)` because it keeps the existing global pattern (so reset/tests stay simple), avoids `cache-related overhead` and `GC` complications, and ensures the fast path has no lock overhead—just a quick None check.

##### For all contributions
- [ ] I've added unit tests and made sure they pass locally.

##### For new estimators
None

`pytest` pass locally
image

lint pass locally
image

### DIFF
--- a/src/sktime_mcp/composition/validator.py
+++ b/src/sktime_mcp/composition/validator.py
@@ -15,6 +15,7 @@ This prevents invalid pipelines at planning time.
 """
 
 import logging
+import threading
 from dataclasses import dataclass, field
 from enum import Enum
 from typing import Any, Optional
@@ -400,11 +401,14 @@ class CompositionValidator:
 
 # Singleton instance
 _validator_instance: Optional[CompositionValidator] = None
+_validator_lock = threading.Lock()
 
 
 def get_composition_validator() -> CompositionValidator:
     """Get the singleton composition validator instance."""
     global _validator_instance
     if _validator_instance is None:
-        _validator_instance = CompositionValidator()
+        with _validator_lock:
+            if _validator_instance is None:
+                _validator_instance = CompositionValidator()
     return _validator_instance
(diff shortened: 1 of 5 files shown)


direkkakkar319-ops:
Hey! @Shashankss1205 , This PR fixes a race condition in all five singleton factory functions. Since f`it_predict_async` uses `loop.run_in_executor(`) to offload work to threads, the existing if None checks could be hit concurrently, creating duplicate instances.

Shashankss1205:
On current main the singleton factories are only ever called from the server's event-loop thread (tools, evaluate, the periodic cleanup); background work receives the already-built objects through closures, and the shared state those workers mutate is locked inside HandleManager (#555) and JobManager. Combined with composition/validator.py no longer existing, there is nothing left for this patch to guard. Closing; if a factory call from a worker thread ever appears, a small double-checked lock including get_job_manager would be welcome as a fresh PR.

PR #85, opened 2026-04-06
fix: replace deprecated asyncio.get_event_loop() with managed scheduler

tarun-227:
## Summary
- Introduce `AsyncScheduler` in `src/sktime_mcp/runtime/async_scheduler.py` that manages a dedicated background thread with its own event loop
- Replace direct `asyncio.get_event_loop()` calls in `fit_predict_async_tool` and `load_data_source_async_tool` with the new scheduler
- Add done-callback logging so unhandled background exceptions are observable
- Add 7 tests covering scheduler behavior and tool-level integration

## Context
Fixes #65. The async tool entrypoints (`fit_predict_async_tool`, `load_data_source_async_tool`) used `asyncio.get_event_loop()` from synchronous code, which is deprecated in Python 3.10+ and emits `DeprecationWarning: There is no current event loop`. The new scheduler avoids this by maintaining a long-lived background loop thread.

## Test plan
- [x] All 7 new tests in `test_async_scheduler.py` pass
- [x] No `DeprecationWarning` about event loops in test output
- [x] All 20 existing tests in `test_core.py` still pass (no regressions)
- [x] All 4 existing tests in `test_async_data_loading.py` still pass
- [x] `ruff check` passes on all changed files

AI was used for basic drafting assistance and refinement; all changes were verified manually.

PR #77, opened 2026-04-05
[BUG]: fixing the async `fit_predict` scheduling to use the running event loop

direkkakkar319-ops:
### What does this implement/fix? Explain your changes.
Ensure `fit_predict_async_tool` schedules background work on the already running event loop via `asyncio.get_running_loop()` and `loop.create_task`, avoiding the deprecated `get_event_loop` + `run_coroutine_threadsafe` path that could leave jobs stuck pending (`src/sktime_mcp/tools/fit_predict.py:160`).

#### Does your contribution introduce a new dependency? If yes, which one?
No new dependencies.

#### What should a reviewer concentrate their feedback on?
Confirm this function is always invoked from an async context with a running loop (per current MCP server flow).
Double-check job status transitions now progress as expected after scheduling.

#### Any other comments?
Tests not added; please let me know if a minimal regression test is desired for async job submission.

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [ ] I've added unit tests and made sure they pass locally.

##### For new estimators
- [ ] I've added the estimator to the online documentation.
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

rant the checks locally 
link: https://github.com/user-attachments/assets/893028dc-7d65-476d-abea-4f280456ad44

PR #105, opened 2026-04-08
fix: fail fast in fit_predict_async for invalid estimator handles

Saloni-0465:
#### Reference Issues/PRs

None (discovered during code review of async tool behavior). Related PRs #77 and #85 fix asyncio scheduling mechanics but do not address this validation gap.

`fit_predict_async_tool` returns `success: true` even when the provided `estimator_handle` does not exist. The previous code caught the exception from `get_info()`, logged a warning, set `estimator_name = "Unknown"`, and proceeded to schedule a background job - which would then immediately fail with "Handle not found".

This means the LLM (or user) gets `success: true` and a `job_id`, wastes time polling `check_job_status`, and only then discovers the handle was invalid.

No.

- Whether failing fast on invalid handles is the right behavior for an async tool (vs. letting the job start and fail gracefully)
- Whether `Exception` is the right catch scope, or if it should be narrowed to `KeyError`

#### Any other comments?

You can verify the fix with:

    from sktime_mcp.tools.fit_predict import fit_predict_async_tool

    result = fit_predict_async_tool("est_nonexistent", "airline", horizon=12)
    assert not result["success"]
    assert "est_nonexistent" in result["error"]

##### For all contributions

### DIFF
--- a/src/sktime_mcp/tools/fit_predict.py
+++ b/src/sktime_mcp/tools/fit_predict.py
@@ -143,9 +143,11 @@ def fit_predict_async_tool(
     try:
         handle_info = executor._handle_manager.get_info(estimator_handle)
         estimator_name = handle_info.estimator_name
-    except Exception as e:
-        logger.warning(f"Could not get estimator name: {e}")
-        estimator_name = "Unknown"
+    except Exception:
+        return {
+            "success": False,
+            "error": f"Unknown estimator handle: {estimator_handle}",
+        }
 
     # Create job
     job_id = job_manager.create_job(


Shashankss1205:
The gap you found is still present on today's main, just in the new tools: fit, predict and evaluate with run_async=True return success: true and a job id for a nonexistent estimator handle, and the job fails a second later with 'Handle not found' (reproduced on f6fb797). This PR itself cannot merge because fit_predict_async_tool was removed in #487, so please rebase and apply the same early return to the `except Exception: estimator_name = 'Unknown'` blocks in tools/fit_predict.py (fit_tool, predict_tool) and tools/evaluate.py, catching the handle manager's specific error rather than bare Exception, plus a small test for each. Happy to merge a tight version of that.

Me-Priyank:
Looks like this needs a rebase, `fit_predict_async_tool` was removed in `493d5a8` (#487).

The gap is still there in its replacements though: `fit_tool` at `fit_predict.py:93-97` and `predict_tool` at `:164-167` both swallow the lookup and create a job anyway. If you do rebase, `get_info` raises `KeyError` specifically, and there's `describe_missing()` which tells evicted apart from unknown, `evaluate.py:113-127` already uses that pattern.

PR #107, opened 2026-04-09
fix: update export_code to generate task-aware examples instead of forecasting-only template

Saloni-0465:
#### Reference Issues/PRs
Relates to #68. Integrates with the dataset selection logic from #76.

This PR makes `export_code_tool(..., include_fit_example=True)` generate task-aware fit/predict examples.

Previously, exported examples always used a forecasting template with `load_airline()`, `fit(y)`, and `predict(fh=fh)`. That works for forecasters, but produces misleading or broken examples for classifiers, regressors, and other estimator types.

This PR resolves the estimator task from the registry and generates:

- forecasting: dataset loader + `fit(y)` + `predict(fh=fh)`
- classification/regression: placeholder `fit(X, y)` + `predict(X_test)` pattern
- other tasks: short pointer to the relevant sktime task API

For forecasting examples, this now reuses the dataset selection logic from #76, so the generated code uses the selected demo dataset loader while still applying the correct forecasting fit/predict pattern.

It also supports pipeline handles by resolving the task from the final pipeline component.

No.

- The integration between `_get_demo_datasets()` from #76 and `_resolve_task()`
- Whether the non-forecasting examples should remain generic placeholders or use concrete datasets
- Whether resolving pipeline task from the final component is the right behavior

#### Any other comments?
This is complementary to #76: that PR improves dataset selection, while this PR fixes the task-type mismatch in generated examples.

#### Validation
- `python -m compileall src/sktime_mcp/tools/codegen.py`

### DIFF
--- a/src/sktime_mcp/tools/codegen.py
+++ b/src/sktime_mcp/tools/codegen.py
@@ -51,6 +51,24 @@ def _is_valid_var_name(var_name: str) -> bool:
     return isinstance(var_name, str) and var_name.isidentifier() and not keyword.iskeyword(var_name)
 
 
+def _resolve_task(
+    estimator_name: str, is_pipeline: bool, params: dict[str, Any]
+) -> str:
+    """Determine the task type for an estimator or pipeline handle."""
+    registry = get_registry()
+    if is_pipeline:
+        components = params.get("components", [])
+        if components:
+            last = registry.get_estimator_by_name(components[-1])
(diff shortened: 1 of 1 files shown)


Shashankss1205:
The export examples are now generated per scitype on main (#534), including pipelines, chosen datasets and loaded models, which covers what this PR set out to do. Closing as already fixed.

Issue #92, opened 2026-04-06
[ENH] Expose existing fit/predict and handle-management tools via MCP server


AnshumanAI:
## Problem

There is a mismatch between implemented tool functions and MCP-exposed server tools.

The following tool functions already exist in code:
- `fit_tool`
- `predict_tool`
- `list_handles_tool`
- `release_handle_tool`

But they are not consistently exposed through MCP server tool registration/dispatch on upstream `main`.

This makes these capabilities unavailable to MCP clients even though backend implementations exist.

## Why this matters

- Creates API drift between runtime capabilities and MCP-visible interface.
- Causes confusion for contributors/users when docs/examples imply availability.
- Prevents direct handle management and separated fit/predict workflows from clients.

## Proposed solution

Update `src/sktime_mcp/server.py` to expose and route:
- `fit`
- `predict`
- `list_handles`
- `release_handle`

Specifically:
1. Add tool schemas in `list_tools()`.
2. Add dispatch branches in `call_tool()`.
3. Ensure `predict` output is JSON-safe (same pattern used by other prediction-returning tools).
4. Add/adjust tests in `tests/test_core.py`:
   - tool exposure in server tool list
   - list/release handle flow
5. Sync `README.md` tool list with actual exposed MCP API.

## Acceptance criteria

- MCP clients can call all 4 tools successfully.
- Tests for these flows pass in CI.
- README reflects the actual server-exposed tool set.

## Context

A previous PR with this change was closed with request to open an issue first:
https://github.com/sktime/sktime-mcp/pull/50#issuecomment-4187392776

AnshumanAI:
@Shashankss1205 Please assign this issue to me.
Thanks.

AnshumanAI:
@tarun-227 I had opened a PR already.
But no worries.

Shashankss1205:
All four tools (fit, predict, list_handles, release_handle) are registered and dispatched on main and are part of the current 26-tool surface. Closing as completed.

PR #45, opened 2026-04-04
feat: add load_model tool for resolving trained models

amruth6002:
*Note: I realize #32 was just assigned to someone else, but I had already finished these changes locally before it was assigned because no  contributor asked to take upon this issue at that time. I figured I would put the PR up just in case having it done is helpful to the project.*
#### Reference Issues/PRs
Closes #32 
#### What does this implement/fix? Explain your changes.
This implements a new `load_model` tool to allow LLMs to restore previously trained estimators from a local directory or URI without having to re-fit the data from scratch.
**Implementation Details:**
- Created `load_model_tool` in `src/sktime_mcp/tools/instantiate.py`.
- Gracefully handles the optional `mlflow` dependency. If `mlflow` is missing, it returns a clear error message instructing the user/LLM to run `pip install sktime[mlflow]` rather than crashing the server.
- Registers the loaded estimator instance dynamically back into the `HandleManager` marking it as fitted, allowing the LLM to immediately follow up with a `predict` call using the newly restored handle.
- Added the tool routing logic to `call_tool` in `server.py`.
#### Does your contribution introduce a new dependency? If yes, which one?
No new hard dependencies. `mlflow` is utilized as an optional extension (`sktime[mlflow]`), strictly matching `sktime`'s existing dependency strategy.

#### Demo Video
[load_model.webm](https://github.com/user-attachments/assets/d36c8be5-8736-4ec5-be95-2bd534db799a)

PR #110, opened 2026-04-09
feat: expose separate fit and predict tools as MCP endpoints

tarun-227:
## Summary
- Register `fit` and `predict` as independent MCP tools in `server.py`
- Enables LLMs to **fit once, predict many times** with different horizons without re-training
- Adds clear tool descriptions guiding LLMs on when to use `fit`/`predict` vs `fit_predict`
- Prediction output is JSON-sanitized (handles Period index objects)

## Why this matters
Currently, `fit_predict` couples training and inference into a single call. If an LLM wants to try 3 different forecast horizons, it must re-fit the model 3 times. With separate `fit` + `predict`, the model trains once and the LLM can predict with `horizon=3`, `horizon=12`, `horizon=24` instantly.

This also enables inspection workflows — fit a model, check `list_handles` to confirm fitted status, then predict.

## Changes
| File | Change |
|------|--------|
| `server.py` | Import `fit_tool`, `predict_tool`; add Tool schemas + dispatch |
| `tests/test_core.py` | 4 new tests covering fit→predict flow, error cases, multi-horizon |

## Test plan
- [x] `test_fit_then_predict` — fit + predict produces results
- [x] `test_predict_without_fit_fails` — unfitted estimator returns error
- [x] `test_fit_then_predict_different_horizons` — one fit, two predicts with horizon=3 and horizon=12
- [x] `test_list_handles_shows_fitted_status` — fitted flag updates correctly
- [x] All 19 existing tests pass (0 regressions)
- [x] `ruff check` clean

Partially addresses #92

AI was used for basic drafting assistance and refinement; all changes were verified manually.

### DIFF
--- a/src/sktime_mcp/server.py
+++ b/src/sktime_mcp/server.py
@@ -30,6 +30,8 @@ from sktime_mcp.tools.describe_estimator import (
 from sktime_mcp.tools.fit_predict import (
     fit_predict_async_tool,
     fit_predict_tool,
+    fit_tool,
+    predict_tool,
 )
 from sktime_mcp.tools.format_tools import (
     auto_format_on_load_tool,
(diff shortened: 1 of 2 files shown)


amruth6002:
hii @tarun-227 

i think load_model tool (#45) solve this, we can fit once and then load that model to predict many times.

tarun-227:
Hi @amruth6002, thanks for the feedback!

`load_model` does enable persistence across server restarts, but the workflow is quite different from what separate `fit`/`predict` tools enable:

| | `save_model` + `load_model` | `fit` + `predict` |
|---|---|---|
| **Requires** | MLflow installed, disk I/O | Nothing extra — in-memory |
| **Steps for 3 horizons** | fit_predict → save_model → load_model → predict (×3) | fit → predict (×3) |
| **Use case** | Persist across sessions | Iterate within a session |

The key scenario: an LLM wants to try `horizon=3`, `horizon=12`, and `horizon=24` on the same fitted model. With `fit_predict` alone, it must re-train 3 times. With `fit` + `predict`, it trains once and predicts instantly — no disk writes, no MLflow dependency.

These are complementary patterns — `save_model`/`load_model` for persistence, `fit`/`predict` for in-session iteration. Both are valuable.

This also aligns with the original request in #92 which specifically asks for `fit` and `predict` as separate MCP tools.

Shashankss1205:
This was the right direction, and it landed on main in #487: fit and predict are now the primary execution tools (the fit_predict macros were removed), with run_async, handle-based inputs and prediction modes. Closing as already implemented.

PR #114, opened 2026-04-10
feat: add detect_seasonality and check_structural_break diagnostic tools (#96)

### LINKED ISSUE #96: [ENH] Two Statistical Diagnostic MCP Tools
Hi @Shashankss1205 @fkiraly and the sktime-mcp team!

While working on my own agentic forecasting project on top of sktime-mcp, I ended up building two custom MCP tools that are purely statistical / preprocessing in nature. I designed them to give an LLM agent clear, actionable signals about the data *before* it decides on the next modeling step — think of them as the agent's "eyes" on the raw series.

I think they could be a nice fit (or at least useful inspiration) for sktime-mcp, so wanted to share the details and get your thoughts!

---

### 1. `detect_seasonality` — Detect cyclic patterns and quantify their strength

Before the agent picks any seasonal model, it should *know* whether a cycle even exists, what period it runs on, and how strong it is.

Under the hood it does a few things worth calling out:

- **Adaptive detrending** — applies first-order differencing only if it reduces the series variance, otherwise uses the raw data. This prevents a strong trend from bleeding into the ACF scores and faking a seasonality signal, without unnecessarily destroying useful structure.
- **FFT-accelerated ACF** — computes the full autocorrelation in `O(N log N)` using NumPy FFT instead of a naive scan, so it stays fast on large series.
- **Dynamic significance threshold** — uses whichever is stricter between the 95% confidence bound (`1.96 / sqrt(N)`) and a hard `0.2` floor. On small datasets the confidence bound dominates; on large ones the `0.2` floor kicks in.
- **Optional frequency awareness** — if the caller passes a known frequency (e.g. `"D"`, `"W"`, `"M"`), known seasonal periods for that frequency get a small scoring bonus, which helps break ties cleanly.

Sample output the agent sees:
```json
{
  "period": 7,
  "strength": 0.78,
  "seasonality_class": "strong",
  "confidence": "high",
  "candidates": [
    {"period": 7, "correlation": 0.78},
    {"period": 14, "correlation": 0.61},
    {"period": 28, "correlation": 0.44}
  ],
  "method": "autocorrelation_detrended_fft",
  "next_action_hint": "deseasonalize_then_stationarity"
}
```

The `next_action_hint` is probably the most useful field for agentic loops — the LLM doesn't have to interpret raw numbers, it just gets a direct nudge on what to do next.

---

### 2. `check_structural_break` — Detect permanent regime changes using CUSUM

This detects if the series has had a sudden, permanent level shift at some point in history (think: a system outage that permanently dropped traffic, or a viral event that permanently spiked sales). Training a model across a structural break silently destroys forecast quality — the agent needs to catch this *before* fitting.

How it works:

- **No differencing (intentional)** — raw data is used. Differencing would collapse a permanent level shift into a single localized blip, destroying the evidence of a regime change.
- **Retrospective global CUSUM** — computes the cumulative sum of mean-centered data. A level shift produces a distinct mountain/valley shape; the break point is the index of the absolute maximum divergence.
- **`k = 0`** — since we analyze the full historical dataset at once (not streaming), no slack parameter is needed to filter real-time noise.
- **Dynamic threshold `h`** — the test statistic (max CUSUM divergence / `std * sqrt(N)`) is compared against `1.358 * (1 + 0.1 * ln(N))`. The `1.358` base comes from the 95% confidence bound of a Brownian Bridge; the log-scaled adjustment prevents long datasets from triggering false positives through random walk drift alone.

Sample output:
```json
{
  "break_detected": true,
  "location": 412,
  "confidence": 0.847,
  "test_stat": 3.91,
  "next_action_hint": "get_dataset_history"
}
```

---

### How these fit the broader roadmap

These two tools are meant to sit *before* the tools in #24 in the reasoning loop, so they complement rather than overlap with what's already being built there:

```
detect_seasonality
       ↓
check_structural_break
       ↓
extract_ts_metadata   ← proposed in #24
       ↓
evaluate_estimator    ← proposed in #24, assigned to @yash-sangwan
```

If this matches the project's scope and direction, I'm happy to open two separate PRs for these. Would love to hear your thoughts before I do!

SarthakB11:
Closes #96

## Summary

Two new MCP tools for pre-modeling statistical diagnostics, following @nXtCyberNet's design from #96:

### `detect_seasonality`
- FFT-accelerated autocorrelation (`O(N log N)`) with adaptive detrending
- Dynamic significance threshold: `max(1.96/sqrt(N), 0.2)`
- Optional frequency-aware scoring (boosts known periods for D/W/M data)
- Returns: `period`, `strength`, `seasonality_class`, `confidence`, `candidates`, `next_action_hint`

### `check_structural_break`
- Retrospective global CUSUM on raw (non-differenced) data
- Dynamic threshold: `1.358 * (1 + 0.1 * ln(N))` (Brownian Bridge 95% bound)
- Returns: `break_detected`, `location`, `confidence`, `test_stat`, `next_action_hint`

Both tools provide `next_action_hint` to guide LLM agents in agentic forecasting workflows.

## Changes

- `src/sktime_mcp/tools/diagnostics.py` — tool implementations (265 lines)
- `src/sktime_mcp/server.py` — tool registration in `list_tools` + `call_tool`
- `tests/test_diagnostics.py` — 12 tests covering both tools (seasonality detection, structural breaks, edge cases, output structure)

## Test plan

- [x] 12 new tests pass (`pytest tests/test_diagnostics.py`)
- [x] 42 existing tests still pass (`pytest tests/`)
- [x] No new dependencies (uses only `numpy` which is already a dep)

### DIFF
--- a/src/sktime_mcp/server.py
+++ b/src/sktime_mcp/server.py
@@ -47,6 +47,10 @@ from sktime_mcp.tools.list_estimators import (
     get_available_tags,
     list_estimators_tool,
 )
+from sktime_mcp.tools.diagnostics import (
+    check_structural_break_tool,
+    detect_seasonality_tool,
+)
 from sktime_mcp.tools.save_model import save_model_tool
 
 # ---------------------------------------------------------------------------
@@ -560,6 +564,44 @@ async def list_tools() -> list[Tool]:
                 "required": ["job_id"],
             },
         ),
+        Tool(
+            name="detect_seasonality",
+            description="Detect cyclic patterns in a time series and quantify their strength. Uses FFT-accelerated autocorrelation with adaptive detrending. Returns period, strength, confidence, candidate periods, and a next_action_hint for the agent.",
+            inputSchema={
+                "type": "object",
+                "properties": {
+                    "data": {
+                        "type": "array",
+                        "items": {"type": "number"},
+                        "description": "The time series values as a flat list of floats",
+                    },
+                    "frequency": {
+                        "type": "string",
+                        "description": "Optional frequency hint (e.g. 'D', 'W', 'M') to boost known seasonal periods",
+                    },
+                    "max_lag": {
+                        "type": "integer",
+                        "description": "Maximum lag to check. Defaults to min(len(data)//2, 200)",
+                    },
+                },
+                "required": ["data"],
+            },
+        ),
+        Tool(
+            name="check_structural_break",
+            description="Detect permanent regime changes (level shifts) in a time series using retrospective CUSUM. Returns whether a break was detected, its location, confidence, and a next_action_hint for the agent.",
+            inputSchema={
+                "type": "object",
+                "properties": {
+                    "data": {
+                        "type": "array",
+                        "items": {"type": "number"},
+                        "description": "The time series values as a flat list of floats",
+                    },
+                },
+                "required": ["data"],
+            },
+        ),
     ]
 
 
@@ -739,6 +781,14 @@ async def call_tool(name: str, arguments: dict[str, Any]) -> list[TextContent]:
             from sktime_mcp.tools.job_tools import cleanup_old_jobs_tool
             result = cleanup_old_jobs_tool(arguments.get("max_age_hours", 24))
 
+        elif name == "detect_seasonality":
+            result = detect_seasonality_tool(
+                arguments["data"],
+                arguments.get("frequency"),
+                arguments.get("max_lag"),
+            )
+        elif name == "check_structural_break":
+            result = check_structural_break_tool(arguments["data"])
         else:
             result = {"error": f"Unknown tool: {name}"}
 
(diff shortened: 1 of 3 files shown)


SarthakB11:
Rebased onto main (import/dispatcher conflicts with the newer job-cleanup tools, resolved by keeping both). Full suite passes locally (215 passed, including the 18 new diagnostics tests). Ready for review whenever you get a chance.

Shashankss1205:
Since this was opened the maintainers consolidated the server to 26 tools (#270) and prefer exposing sktime's own estimators through instantiate/fit/get_fitted_params/call_method over adding task-specific tools; seasonality and change-point detection are already reachable that way (SeasonalityACF from sktime.param_est, detectors in sktime.detection). Closing in line with that direction; if a diagnostics tool is ever warranted, a thin wrapper over those sktime estimators would be the preferred shape.

PR #120, opened 2026-04-10
feat: add predict_interval MCP tool for probabilistic forecasting

FaryalRizwaan:
## Summary

This PR adds a `predict_interval` MCP tool that enables LLM agents 
to generate probabilistic interval forecasts (upper and lower confidence 
bounds) from fitted estimators.

## Motivation

Currently, LLM agents using sktime-mcp can only generate point forecasts 
via `fit_predict`. There is no way to get uncertainty estimates. This is 
a significant gap for real-world agentic workflows where decision-making 
under uncertainty is critical.

## Changes

- `src/sktime_mcp/runtime/executor.py` — added `predict_interval()` method
- `src/sktime_mcp/tools/fit_predict.py` — added `predict_interval_tool()` 
- `src/sktime_mcp/server.py` — registered `predict_interval` as MCP tool
- `tests/test_predict_interval.py` — added 3 tests

## Example Usage

```json
{
  "estimator_handle": "est_abc123",
  "horizon": 12,
  "coverage": 0.9
}
```

Returns:
```json
{
  "success": true,
  "intervals": {
    "1": {"lower": 400.1, "upper": 500.3},
    "2": {"lower": 410.2, "upper": 510.4}
  },
  "coverage": 0.9,
  "horizon": 12
}
```

## Testing

All 3 new tests pass. Full test suite: 44/45 passing.
The 1 pre-existing failure (`test_cleanup_old_jobs`) is unrelated to 
this PR.

## Notes

- Only works with estimators tagged `capability:pred_int=true`
- Compatible estimators can be discovered via:
  `list_estimators(tags={"capability:pred_int": true})`
- Tested with ARIMA on airline dataset


CODEBASE.md
```diff
@@ -0,0 +1,205 @@
+# sktime-mcp Codebase Overview
```
pranavvp16: Seems this is an unnecessary file which is commited

add_method.py
```diff
@@ -0,0 +1,32 @@
+from pathlib import Path
```
pranavvp16: Again some kind of file used by LLM to test the tools

pranavvp16 (changes requested):

FaryalRizwaan:
Fixed in latest push. Both files have been removed.

Shashankss1205:
Interval forecasts are now available on main through the consolidated predict tool with mode='predict_interval' (and predict_quantiles) and a coverage parameter; the output format was tightened in #536. A separate predict_interval tool would duplicate that surface (#270). Closing as already covered.

PR #121, opened 2026-04-11
[ENH] add offset pagination to list_estimators tool

### LINKED ISSUE #112: [ENH] list_estimators returns only 50 results with no indication of total count or pagination support
# Problem:
When calling list_estimators(task="forecasting"), only 50 results are returned out of 128 total forecasting estimators in sktime 0.40.1. The response gives no indication that results are truncated or that more estimators exist beyond the default limit. Requesting all 128 at once causes a timeout with no fallback.

# Proposed Solution:
Add offset and total_count fields to the list_estimators tool response, and support an offset parameter for pagination:

list_estimators(task="forecasting", limit=50, offset=0)
 Returns: {estimators: [...], total: 128, offset: 0, limit: 50}

This allows clients to paginate through all estimators in batches without timeouts, and always know the full count upfront.

# Alternative Solutions Considered:
Raising the default limit  but this causes timeouts for large task categories like forecasting (128 estimators). Pagination is a cleaner solution that scales as sktime grows.

# Additional context:
Discovered while using the MCP with Claude Desktop. The missing total count is especially confusing for LLM clients that reason about available estimators  they may incorrectly assume sktime only has 50 forecasters when the full library has 128+.

# Environment
- sktime-mcp version: 0.1.0
- sktime version: 0.40.1

NestroyMusoke:
Closes #112

## Summary
`list_estimators` previously returned only 50 results with no way to 
access the remaining estimators. For forecasting alone, sktime 0.40.1 
has 128 estimators but only 50 were reachable.

## Changes
- Added `offset` parameter (default: 0) to `list_estimators_tool`
- Added `has_more` field to response so clients know more results exist
- `offset` and `limit` are echoed back in every response
- Added offset validation (rejects negative values)
- Existing calls with no `offset` are fully backwards compatible

## Example
```python
# Page 1
list_estimators(task="forecasting", limit=50, offset=0)
# Returns: {total: 128, count: 50, has_more: True, offset: 0}

# Page 2  
list_estimators(task="forecasting", limit=50, offset=50)
# Returns: {total: 128, count: 50, has_more: True, offset: 50}
```

### DIFF
--- a/src/sktime_mcp/tools/list_estimators.py
+++ b/src/sktime_mcp/tools/list_estimators.py
@@ -1,6 +1,5 @@
 """
 list_estimators tool for sktime MCP.
-
 Discovers estimators by task type and capability tags.
 """
 
@@ -13,6 +12,7 @@ def list_estimators_tool(
     task: Optional[str] = None,
     tags: Optional[dict[str, Any]] = None,
     limit: int = 50,
+    offset: int = 0,
 ) -> dict[str, Any]:
     """
     Discover sktime estimators by task type and capability tags.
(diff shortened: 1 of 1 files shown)


NestroyMusoke:
Closing in favour of cleaner PR #113 which contains only the pagination changes.

PR #128, opened 2026-04-11
fix: include job_id in all failure paths of fit_predict_async

Saloni-0465:
## Reference Issues/PRs
None.

---

## What does this implement/fix?

`fit_predict_async` now always includes `job_id` in the returned response across all paths:
- dataset load failure
- fit failure
- predict failure
- success

Previously, some early failure paths returned only an error dict without `job_id`, even though the job had already been created. This made it impossible for clients to track or inspect failed async jobs.

This change ensures consistency in the async job lifecycle - every response now includes `job_id`.

---

## Does your contribution introduce a new dependency?
No.

---

- Return shapes from `Executor.fit_predict_async`
- Consistency of `job_id` across all execution paths

---

## Any other comments?
None.

---

### For all contributions
- [x] I've added myself to the list of contributors (if applicable)
- [x] I've added unit tests and made sure they pass locally

---

### Title

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -333,7 +333,7 @@ class Executor:
                         status=JobStatus.FAILED,
                         errors=[f"Failed to load dataset: {data_result.get('error')}"],
                     )
-                    return data_result
+                    return {**data_result, "job_id": job_id}
 
                 y = data_result["data"]
                 X = data_result.get("exog")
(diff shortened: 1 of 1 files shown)


direkkakkar319-ops:
hi @Saloni-0465 i saw that you lint and pytests  are failing this is due to problem in the main branch i have fixed both problem in my recent prs

Saloni-0465:
Got it @direkkakkar319-ops - I’ll rebase main once your CI fixes are in so this PR picks them up.

Shashankss1205:
fit_predict_async was removed in the refactor (#487); the replacement fit/predict/evaluate with run_async=True return the job_id up front, and every failure path inside the background coroutine is recorded on the job itself via update_job(status=FAILED, errors=...), so the dict the coroutine returns is no longer client-visible. Closing as obsolete.

PR #130, opened 2026-04-11
fix: failing py-test in main 

### LINKED ISSUE #132: CI check failure
## Description
**there is problem in main branch so the CI Checks related to these are failing:**
1. lint
2. pytest 
3. doc

direkkakkar319-ops:
fixes  #132

in main branch on test was failing 
Screenshot 2026-04-11 231842

after fixing the test 
Screenshot 2026-04-11 231922

- The test in `test_evaluate.py` expected `cv_folds=2`to produce exactly 2 evaluation results.
- In the old code in `evaluate.py` , `cv_folds` was not used to set an exact number of folds. It was only used inside a heuristic for initial_window .
- ExpandingWindowSplitter then generated as many valid splits as the dataset allowed, which happened to be 4 for the airline data.
- So the test failed on assert `len(result["results"]) == 2` because the actual result length was 4 .
What We Fixed

- In `evaluate.py` , we changed the fold logic so `cv_folds` directly controls how many splits are run.
- We now compute:
1. `folds_to_run = min(cv_folds, n - 1)`
2. `initial_window = n - folds_to_run`
3. With `fh=[1]` and `step_length=1` , that makes the splitter return exactly the requested number of folds when possible.

### DIFF
--- a/src/sktime_mcp/tools/evaluate.py
+++ b/src/sktime_mcp/tools/evaluate.py
@@ -8,7 +8,7 @@ import logging
 from typing import Any
 
 from sktime.forecasting.model_evaluation import evaluate
-from sktime.forecasting.model_selection import ExpandingWindowSplitter
+from sktime.split import ExpandingWindowSplitter
 
 from sktime_mcp.runtime.executor import get_executor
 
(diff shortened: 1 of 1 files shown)


Shashankss1205:
The evaluate path was rewritten since April (_run_evaluate in runtime/executor.py now makes cv_folds control the exact number of folds, validates the range, and imports the splitter from sktime.split), so the behaviour you proposed is already on main. Closing as superseded.

PR #133, opened 2026-04-11
fix: restrict auto-format to datetime-like indices to prevent data co…

Saloni-0465:
#### Reference Issues/PRs

N/A

---

Before: When formatting a data handle, the code tried to guess a calendar frequency and fill gaps using pd.date_range and reindex for every kind of index. That is only safe for real date/time style indexes. For simple row indexes (like RangeIndex or plain integers), it could change how many rows you have, add missing values, or otherwise mess up the series while still reporting success.

After: That calendar step only runs for proper date indexes (DatetimeIndex) and period indexes (PeriodIndex). For anything else, we skip that step and record why in changes_made (calendar_gap_fill_skipped and calendar_gap_fill_reason).

We also catch errors from pd.infer_freq when pandas refuses to infer (for example, very few dates).

We fixed a crash when saving metadata: some indexes don’t have a .freq attribute, so we use getattr(y.index, "freq", None) instead of assuming it exists.

Tests: tests/test_format_data_handle.py.

---

No.

---

- Correctness of the DatetimeIndex path vs previous behavior (gap-fill and changes_made)  
- Whether PeriodIndex handling is acceptable or should be narrowed further  
- Whether calendar_gap_fill_* fields in changes_made are the right contract for callers  

---

#### Any other comments?

To verify locally:  
`python -m pytest tests/test_format_data_handle.py -v`

---

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).

##### For new estimators

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -50,6 +50,86 @@ def _get_demo_datasets() -> dict:
     return _DEMO_DATASETS
 
 
+def _apply_calendar_frequency_and_gap_fill(
+    y: pd.Series,
+    X: pd.DataFrame | None,
+    changes_made: dict[str, Any],
+) -> tuple[pd.Series, pd.DataFrame | None]:
+    """
+    Infer frequency and reindex to a full calendar range when safe.
+
+    Only ``DatetimeIndex`` and ``PeriodIndex`` support this path. Other index
+    types (e.g. ``RangeIndex``) are left unchanged: inferring a daily frequency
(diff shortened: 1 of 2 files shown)


Shashankss1205:
This was a real bug, and it was fixed on main in #391: format_data_handle now only infers/reindexes for DatetimeIndex/PeriodIndex and uses getattr(index, 'freq', None); re-checked with a RangeIndex series on current main and the rows are preserved. The branch would also reintroduce the silent 'D' fallback that #553 just removed. Closing as already fixed.

PR #147, opened 2026-04-12
fix: cancel_job() now actually stops running background jobs (#136)

### LINKED ISSUE #136: [BUG] cancel_job() does not actually stop running background jobs cancelled status gets overwritten to completed
**Describe the bug**
When calling cancel_job_tool() on a running background job (started via fit_predict_async_tool), the cancel_job() method in JobManager only changes the job's status label to CANCELLED but does not actually stop the running coroutine. The asyncio.Future returned by asyncio.run_coroutine_threadsafe() in fit_predict_async_tool() is discarded and never stored, so there is no way to cancel the actual computation. The background coroutine continues to run, and when it finishes, it overwrites the CANCELLED status back to COMPLETED.

**To Reproduce**

```python
from sktime_mcp.runtime.jobs import JobManager, JobStatus

job_manager = JobManager()

# Create and start a job
job_id = job_manager.create_job(
    job_type="fit_predict",
    estimator_handle="test_handle",
    estimator_name="ARIMA",
    dataset_name="airline",
    total_steps=3,
)
job_manager.update_job(job_id, status=JobStatus.RUNNING)

# Cancel the job
result = job_manager.cancel_job(job_id)
print(f"Cancel returned: {result}")  # True
print(f"Status after cancel: {job_manager.get_job(job_id).status.value}")  # "cancelled"

# Simulate the background coroutine finishing
job_manager.update_job(
    job_id,
    status=JobStatus.COMPLETED,
    completed_steps=3,
    current_step="Completed",
    result={"predictions": {1: 450.2, 2: 460.5}},
)
print(f"Status after coroutine finishes: {job_manager.get_job(job_id).status.value}")
# Output: "completed" the CANCELLED status was silently overwritten.
```
Also, in fit_predict_async_tool() (src/sktime_mcp/tools/fit_predict.py, line 174), the return value of asyncio.run_coroutine_threadsafe() is discarded.

```
coro = executor.fit_predict_async(estimator_handle, dataset, horizon, job_id)
asyncio.run_coroutine_threadsafe(coro, loop)  # Future is thrown away, can never be cancelled
```

**Expected behavior**
When a user cancels a job via cancel_job_tool(), the background computation should actually stop (or at least the coroutine should check for cancellation between steps and exit early). The CANCELLED status should not be silently overwritten to COMPLETED by the still-running coroutine.

**Additional context**
There are three related issues here:

1. Future is discarded - In fit_predict_async_tool() (fit_predict.py line 174), asyncio.run_coroutine_threadsafe(coro, loop) returns a Future that is never stored. Without a reference to this Future, there is no way to call .cancel() on it.

2. No cancellation check in coroutine - The fit_predict_async() method in executor.py never checks job.status between its three steps (load data → fit → predict). Even after cancel is called, the coroutine proceeds through all steps.

3. Status overwrite - update_job() in jobs.py does not guard against overwriting a CANCELLED status. When the coroutine finishes, it calls update_job(status=JobStatus.COMPLETED), which silently overwrites the cancel.

The same issue also affects load_data_source_async_tool() in data_tools.py.

**Versions**
```
System:
    python: 3.10+
    machine: macOS / Ubuntu CI environment

Python dependencies:
    sktime: 0.28.0 (or latest main)
    mcp: >= 1.0.0
```

Abhishek9639:
### Reference Issues/PRs

Fixes #136

`cancel_job()` was only changing the status label to CANCELLED but wasn't actually stopping the background coroutine. The coroutine would keep running and silently overwrite the CANCELLED status back to COMPLETED when it finished.

There were three things causing this:

1. The `Future` returned by `asyncio.run_coroutine_threadsafe()` was being thrown away in both `fit_predict_async_tool()` and `load_data_source_async_tool()`, so there was no way to actually cancel it.

2. The async coroutines (`fit_predict_async` and `load_data_source_async`) never checked if the job had been cancelled between their steps they just ran straight through load → fit → predict regardless.

3. `update_job()` had no guard against overwriting a CANCELLED status, so the still-running coroutine could silently set it back to COMPLETED.

This PR fixes all three:
- Stores the `Future` via a new `register_future()` method and calls `future.cancel()` in `cancel_job()`
- Adds cancellation checks between each step in both async coroutines so they exit early
- Adds a guard in `update_job()` that rejects status changes from CANCELLED

No.

- The guard logic in `update_job()` it currently blocks any status change away from CANCELLED. This felt like the right approach since a cancelled job should stay cancelled, but open to feedback if FAILED should still be allowed to overwrite CANCELLED.
- The cancellation check pattern in the async coroutines checks happen at the yield points between steps. During a blocking `run_in_executor` call (e.g., while fitting a model), the cancellation won't take effect until that step finishes. This is a limitation of cooperative cancellation but seemed like a reasonable tradeoff.

### Any other comments?

The same issue also affected `load_data_source_async_tool()` in `data_tools.py` as noted in the issue, so I applied the fix there too.

Three new tests added:
- `test_cancel_prevents_overwrite` - verifies the exact reproduction scenario from the issue
- `test_cancel_job_stores_and_cancels_future` - verifies `future.cancel()` is called
- `test_async_fit_predict_respects_cancellation` - end-to-end test for early exit

#### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/CONTRIBUTORS.md).

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -266,6 +266,11 @@ class Executor:
             )
             await asyncio.sleep(0.01)  # Yield control to event loop
 
+            # Check if job was cancelled before proceeding
+            job = self._job_manager.get_job(job_id)
+            if job and job.status == JobStatus.CANCELLED:
+                return {"success": False, "error": "Job cancelled", "job_id": job_id}
+
             data_result = self.load_dataset(dataset)
             if not data_result["success"]:
                 self._job_manager.update_job(
@@ -284,6 +289,11 @@ class Executor:
                 job_id, completed_steps=1, current_step=f"Fitting {estimator_name} on {dataset}..."
             )
             await asyncio.sleep(0.01)  # Yield control
+
+            # Check if job was cancelled before proceeding
+            job = self._job_manager.get_job(job_id)
+            if job and job.status == JobStatus.CANCELLED:
+                return {"success": False, "error": "Job cancelled", "job_id": job_id}
 
             # Run fit in executor to avoid blocking
             loop = asyncio.get_event_loop()
@@ -306,6 +316,11 @@ class Executor:
                 current_step=f"Generating predictions (horizon={horizon})...",
             )
             await asyncio.sleep(0.01)  # Yield control
+
+            # Check if job was cancelled before proceeding
+            job = self._job_manager.get_job(job_id)
+            if job and job.status == JobStatus.CANCELLED:
+                return {"success": False, "error": "Job cancelled", "job_id": job_id}
 
             # Run predict in executor
             predict_result = await loop.run_in_executor(
@@ -611,6 +626,11 @@ class Executor:
             )
             await asyncio.sleep(0.01)
 
+            # Check if job was cancelled
+            job = self._job_manager.get_job(job_id)
+            if job and job.status == JobStatus.CANCELLED:
+                return {"success": False, "error": "Job cancelled", "job_id": job_id}
+
             from sktime_mcp.data import DataSourceRegistry
 
             loop = asyncio.get_event_loop()
@@ -622,6 +642,11 @@ class Executor:
                 job_id, completed_steps=1, current_step="Validating data..."
             )
             await asyncio.sleep(0.01)
+
+            # Check if job was cancelled
+            job = self._job_manager.get_job(job_id)
+            if job and job.status == JobStatus.CANCELLED:
+                return {"success": False, "error": "Job cancelled", "job_id": job_id}
 
             is_valid, validation_report = adapter.validate(data)
             if not is_valid:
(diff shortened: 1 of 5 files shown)


Abhishek9639:
Greetings @Shashankss1205,
Opened a fix for this would appreciate a review when you get a chance. 
Thanks

Abhishek9639:
Greetings @Shashankss1205,
Just a heads up the CI failures here are all pre-existing on main (lint issues in evaluate.py/instantiate.py/server.py, the test_evaluate.py assertion failure, and the Python 3.9 incompatibility with the mcp package). None of them are related to the changes in this PR. 
I verified locally that all files I touched pass black, ruff, and pytest cleanly.
Thanks

Shashankss1205:
Your analysis in #136 was correct on all three root causes. They were fixed on main in #513: JobManager.register_task keeps the asyncio task, cancel_job calls task.cancel() on it, and update_job ignores late status updates on a cancelled job, with tests in tests/test_job_cancellation.py and tests/test_background_jobs.py. Closing as superseded; #136 is closed as fixed.

PR #157, opened 2026-04-12
[ENH] Implement pagination for list_available_data and list_data_sources tools

### LINKED ISSUE #153: [ENH] Implement pagination for list_available_data and list_data_sources tools
The `list_available_data_tool` and `list_data_sources_tool` currently return all results in a single response with no pagination support. This is inconsistent with `list_estimators_tool` which was recently updated in #113 to support `limit` and `offset` parameters. As the number of datasets and data handles grows, an unpaginated response can exceed token limits and degrade MCP performance.

Add `limit: int = 50` and `offset: int = 0` parameters to `list_available_data_tool` in `list_available_data.py` and `list_data_sources_tool` in `data_tools.py`, following the exact same pattern introduced in `list_estimators_tool`. The return dictionary should be extended to include `total`, `count`, `has_more`, `offset`, and `limit` fields to make all list tools consistent in their response schema.

Returning all results at once (current behaviour) is the alternative, but this risks very large payloads when many handles or sources are registered. Adding a `search` filter was also considered, but pagination is a simpler and more composable solution that works alongside existing filters.

**Additional context**
This change makes all `list_*` tools in `sktime-mcp` follow a single consistent interface, which is important for LLM agents that rely on predictable response schemas. The implementation would mirror what was already merged in #113, so the pattern is proven and tested.

rupeshca007:
## Summary

Implements pagination (`limit` / `offset`) for `list_available_data_tool` and `list_data_sources_tool`, closing the consistency gap introduced when `list_estimators_tool` was updated in #113.

### `src/sktime_mcp/tools/list_available_data.py`
- Added `limit: int = 50` and `offset: int = 0` parameters to `list_available_data_tool`
- - Return dict now includes `total`, `count`, `has_more`, `offset`, `limit` fields
- - Negative `offset` returns a graceful error response
- - Items are combined into a flat list for unified pagination, then split back into `system_demos` / `active_handles` for backward-compatible response shape
### `src/sktime_mcp/tools/data_tools.py`
- Added `limit: int = 50` and `offset: int = 0` parameters to `list_data_sources_tool`
- - Return dict now includes `total`, `count`, `has_more`, `offset`, `limit` fields
- - Descriptions dict only covers the paged sources, not all adapters
- - Negative `offset` returns a graceful error response
### `tests/test_pagination.py` (new file)
- 14 pytest tests covering both tools
- - Tests include: default fields, limit enforcement, offset-beyond-total, negative offset error, `has_more` flag, count/descriptions consistency
## Motivation

Fixes #153. All `list_*` tools should follow the same pagination interface for predictable LLM agent usage. This mirrors the pattern already merged in #113.

## Testing

```
pytest tests/test_pagination.py -v
# 14 passed
```


Shashankss1205:
already exists in the main

PR #159, opened 2026-04-12
fix auto_format_on_load KeyError when called without arguments

### LINKED ISSUE #155: [BUG] auto_format_on_load crashes with KeyError when called without arguments
**Describe the bug**

Calling the `auto_format_on_load` tool without passing `enabled` throws a `KeyError` instead of using the default value. The schema says `enabled` is optional (default `True`), but `server.py` line 686 does `arguments["enabled"]` instead of using `.get()`.

**To Reproduce**

```
from sktime_mcp.server import call_tool
import asyncio

result = asyncio.run(call_tool('auto_format_on_load', {}))
print(result)
```

Traceback from server log:

```
File "src/sktime_mcp/server.py", line 686, in call_tool
    result = auto_format_on_load_tool(arguments["enabled"])
KeyError: 'enabled'
```

**Expected behavior**
Should default to enabled=True and return a success response, same as how other optional params like limit, horizon, etc. are handled in the server.

Additional context

Looks like it was just missed every other optional param in call_tool uses arguments.get(...) but this one doesn't.

**Versions**
```
System:
    python: 3.13.7
   machine: Windows-11-10.0.26200-SP0

Python dependencies:
       sktime: 0.40.1
      sklearn: 1.7.2
        numpy: 2.3.5
       pandas: 2.3.3
     pmdarima: 2.1.1
  statsmodels: 0.14.6
```

krishnachaudhary18:
#### Reference Issues/PRs

Fixes #155

`auto_format_on_load` was crashing with `KeyError: 'enabled'` when called without the optional `enabled` argument. 

The issue was on line 686 of `server.py` it used `arguments["enabled"]` (hard key access) instead of `arguments.get("enabled", True)`. This didn't match how every other optional param is handled in `call_tool`.

Changed it to use `.get()` with the default value `True`, which matches the tool's JSON schema.

No.

It's a one-line fix just verify the default value `True` matches what the tool schema and `auto_format_on_load_tool` function signature expect (it does).

#### Any other comments?

Nothing else, straightforward fix.

##### For all contributions


krishnachaudhary18:
Hi @Shashankss1205,
Could you please review this when you get a chance? It's a small one-line fix. 
Thanks

Shashankss1205:
It says file changed is 0. Can you please check this??

PR #162, opened 2026-04-12
[ENH] Implement evaluate_estimator_async_tool for background cross-validation

### LINKED ISSUE #158: [ENH] Implement evaluate_estimator_async_tool for background cross-validation
Currently, `evaluate_estimator_tool` performs synchronous cross-validation using `sktime.forecasting.model_evaluation.evaluate`. Cross-validation is inherently slow, especially for larger datasets, more folds, or computationally heavy foundation models like the new `Chronos2Forecaster`. Because it runs synchronously, it blocks the MCP server and can lead to timeout errors in LLM clients.

Implement an `evaluate_estimator_async_tool` in `src/sktime_mcp/tools/evaluate.py`. This should follow the exact same architectural pattern established by `fit_predict_async_tool` and `load_data_source_async_tool`:
1. Use `get_job_manager().create_job("evaluate", ...)` to register a background task.
2. 2. Schedule a non-blocking coroutine on the executor's event loop to run the evaluation.
3. 3. Return a `job_id` immediately so the LLM agent can monitor progress via `check_job_status`.
Adding an `n_jobs` parameter to the synchronous `evaluate` call could speed up the CV locally, but it does not solve the fundamental issue of the MCP server blocking the LLM client on long tasks. The async jobs pattern is already established in the codebase and is the right solution here.

**Additional context**
This is particularly important as the project moves towards evaluating complex models and processing large agricultural datasets. Those evaluations can be time-consuming, and an async tool provides a much better developer and agentic UX. This is part of my ESoC 2026 application process.

rupeshca007:
## Summary

Adds `evaluate_estimator_async_tool` to allow long-running cross-validation tasks to execute in the background without blocking the MCP server. This closes the consistency gap and directly mirrors the behavior of the recently added `fit_predict_async_tool`.

### `src/sktime_mcp/runtime/executor.py`
- Added an `evaluate_async` method to the `Executor` class.
- Leveraged `_job_manager` to track `RUNNING`, `FAILED`, and `COMPLETED` states exactly like other async jobs.
- Offloaded the synchronous model evaluation logic to `loop.run_in_executor` to prevent blocking the main event loop.

### `src/sktime_mcp/tools/evaluate.py`
- Implemented `evaluate_estimator_async_tool` parallel to the synchronous tool.
- Automatically handles new event loop initialization if running in a thread.

### `src/sktime_mcp/server.py` & `src/sktime_mcp/tools/__init__.py`
- Registered the new `evaluate_estimator_async_tool` with the MCP server schema.
- Added missing `evaluate_estimator_tool` export to `__init__.py`.

### `tests/test_evaluate.py`
- Added comprehensive unit tests that test the full async loop, including polling the job status manager to wait for a `COMPLETED` flag.

## Motivation

Fixes #158. The synchronous cross-validation blocks the server, which leads to timeouts for LLM clients evaluating heavy models, especially foundation models like `Chronos2Forecaster`.

## Testing

```bash
pytest tests/test_evaluate.py -v


Shashankss1205:
Background evaluation was a real need (#158). It has since been implemented following the project's convention of a run_async flag rather than separate *_async tools: evaluate(..., run_async=True) returns a job_id that check_job_status polls, backed by Executor.evaluate_async. Since the feature exists in that form and evaluate_estimator_tool was renamed, closing this PR; #158 is closed as implemented.

Issue #35, opened 2026-04-03
[TESTING] Call for Feedback: Help us test and refine sktime-mcp!


Shashankss1205:
### Welcome Early Adopters! 
The `sktime-mcp` project is still in early development, and we need your help to improve our tools, refine their descriptions, and fix bugs. Your feedback is crucial for making this the most effective time-series co-pilot for AI agents.

### How you can help:
1.  **Run the Server**: Try connecting `sktime-mcp` to different clients (Claude Desktop, Gemini, Vercel AI SDK, etc.).
2.  **Test Core Workflows**:
    *   **Forecasting**: Try instantiating and fitting an `ARIMA` or `Prophet` model on the built-in demo datasets (like `airline` or `sunspots`).
    *   **Custom Data**: Use `load_data_source` to load your own CSVs or databases and run a `fit_predict_with_data` workflow.
    *   **Background Jobs**: Test the `fit_predict_async` capability for long-running training tasks.
3.  **Check Tool Clarity**: Are the descriptions of the tools clear enough for the LLM to understand when and how to use them? If you see an LLM hallucinating a parameter or choosing the wrong tool, let us know!

### What to include in your feedback:
*   **Your Setup**: Which MCP Client and LLM were you using?
*   **What Worked**: Any tools that felt particularly powerful or intuitive?
*   **What Didn't**: Any bugs, cryptic error messages, or confusing "handles"?
*   **Suggestions**: Are there any "quality of life" improvements that would make your analysis workflow smoother?

### Please share your thoughts below! 💬
We are actively iterating, and your input will directly influence our next development cycle. 

Thank you for helping us build the bridge between sktime and LLMs!

Sumit-ai-dev:
Hey @Shashankss1205 . Thanks for opening this up, testing this has been pretty smooth so far. Here are my notes.

## Setup

Instead of a standard desktop app, I built a custom script bridging the MCP `stdio_client` directly to **Gemini 2.5 Flash** (via the official Google GenAI SDK). This let me test exactly how an LLM natively chains your tools together.

## What worked

- **Forecasting**  Tested `ARIMA(1,1,0)` and `ExponentialSmoothing` on the `airline` dataset. The `instantiate` → `fit_predict` flow worked cleanly.
- **Background jobs & Tool chaining** : I asked Gemini to run a background job. It successfully chained `instantiate_estimator` → `fit_predict_async` → `check_job_status` entirely on its own without any human intervention. The async implementation is solid.
[...]

PR #169, opened 2026-04-13
fix: validate params_list length matches components in instantiate_pi…

### LINKED ISSUE #168: [BUG]  instantiate_pipeline silently accepts params_list longer than components
**Describe the bug**
When calling `instantiate_pipeline` with a `params_list` that has more entries than `components`, no error is raised. The extra parameter dicts get validated with `comp_name=None` and silently passed to the executor  so invalid input passes through without any warning.

**To Reproduce**
```python
# 2 components but 3 param dicts. should error, but doesn't
instantiate_pipeline(
    components=["Detrender", "ARIMA"],
    params_list=[{}, {}, {"order": [1, 1, 1]}]
)
```

**Expected behavior**
Should return a clear error saying `params_list` length must match `components` length instead of silently accepting mismatched input.

**Additional context**
The bug is in `src/sktime_mcp/tools/instantiate.py` around line 195. The loop iterates through all items in `params_list` without first checking if its length exceeds `components`.

I will be opening a PR for this shortly to fix is a length check before the loop.

cc @Shashankss1205 

**Versions**
N/A. this is a logic bug in sktime-mcp, not sktime itself.

pyarchana:
#### Reference Issues/PRs
fixes #168 

`instantiate_pipeline` was silently accepting a `params_list` longer than `components`  the extra param dicts got validated with `comp_name=None` and passed to the executor without any error.

So i added a length check in `src/sktime_mcp/tools/instantiate.py` before the validation loop so that mismatched lengths return a clear error message instead of passing through silently.

no.

- The length check added in `instantiate.py` around line 195
- The error message format that is consistent with existing error messages in the same file

#### Any other comments?

This bug was originally flagged in issue #35. Raising a dedicated fix since it's a clear 
validation gap, invalid input was passing through silently to the executor.

#### PR checklist

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1,9 +1,9 @@
-import os
 import sys
 from datetime import datetime
+from pathlib import Path
 
 # Path setup
-sys.path.insert(0, os.path.abspath("../../src"))
+sys.path.insert(0, str(Path("../../src").resolve()))
 
 # Project information
 project = "sktime-mcp"
--- a/src/sktime_mcp/tools/instantiate.py
+++ b/src/sktime_mcp/tools/instantiate.py
@@ -192,8 +192,17 @@ def instantiate_pipeline_tool(
                 ),
             }
 
+        if len(params_list) != len(components):
+            return {
+                "success": False,
+                "error": (
+                    f"'params_list' length ({len(params_list)}) must match "
+                    f"'components' length ({len(components)})"
+                ),
+            }
+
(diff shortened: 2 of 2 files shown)


Shashankss1205:
Since the PR was opened, instantiate_pipeline was replaced by the unified instantiate tool, which takes one sktime craft() spec, so a separate params_list no longer exists and the length mismatch cannot occur. #168 is closed. Closing this PR as obsolete.

PR #177, opened 2026-04-13
[BUG] Propagate coverage parameter through fit_predict_async and executor predict chain

### LINKED ISSUE #176: [BUG] fit_predict_async silently drops coverage parameter - prediction intervals not returned for async workflows
## Summary

`fit_predict_async_tool` and `executor.fit_predict_async` are missing the `coverage` parameter that `fit_predict_tool` already supports. This creates a silent asymmetry: users who request prediction intervals via the async tool receive point forecasts only, with no error or warning.

Worse, `server.py` already passes `coverage=arguments.get("coverage")` to `fit_predict_async_tool`, but the tool's signature does not accept it - meaning **any call to `fit_predict_async` raises a `TypeError`**, even when no coverage is requested.

## Evidence

**`server.py` (already calls with coverage):**
```python
elif name == "fit_predict_async":
    result = fit_predict_async_tool(
        arguments["estimator_handle"],
        arguments["dataset"],
        arguments.get("horizon", 12),
        coverage=arguments.get("coverage"),  # <- passed but not accepted!
    )
```

**`fit_predict_async_tool` signature (missing coverage):**
```python
def fit_predict_async_tool(
    estimator_handle: str,
    dataset: str,
    horizon: int = 12,          # <- no coverage here
) -> dict[str, Any]:
```

**`executor.fit_predict_async` (missing coverage):**
```python
async def fit_predict_async(
    self,
    handle_id: str,
    dataset: str,
    horizon: int = 12,
    job_id: Optional[str] = None,   # <- no coverage here
) -> dict[str, Any]:
```

**`executor.predict` (missing coverage):**
```python
def predict(
    self,
    handle_id: str,
    fh=None,
    X=None,              # <- no coverage, never calls predict_interval()
) -> dict[str, Any]:
```

## Broken Call Chain

```
fit_predict_async (MCP tool call)
  -> fit_predict_async_tool()    <- TypeError: unexpected keyword argument 'coverage'
  -> executor.fit_predict_async() <- no coverage arg
  -> executor.predict()           <- never calls predict_interval()
```

## Steps to Reproduce

1. Instantiate an estimator that supports prediction intervals (e.g. `ExponentialSmoothing`)
2. 2. Call `fit_predict_async` with any `coverage` value (e.g. `0.9`)
3. 3. Observe: `TypeError: fit_predict_async_tool() got an unexpected keyword argument 'coverage'`
## Expected Behavior

`fit_predict_async` should accept `coverage` and return `prediction_intervals` in the completed job result, exactly matching the behavior of `fit_predict`.

## Suggested Fix

Propagate `coverage` through the full async call chain:
- `fit_predict_async_tool(coverage=None)` -> `executor.fit_predict_async(coverage=None)` -> `executor.predict(coverage=None)` -> call `predict_interval()` when coverage is set
## Environment

- Affects all platforms
- - Reproducible on current `main` branch
Happy to submit a PR if the approach is confirmed.

rupeshca007:
## Summary

Fixes #176.

`server.py` already passes `coverage=arguments.get("coverage")` to `fit_predict_async_tool`, but the tool's signature did not accept it - causing a `TypeError` on every `fit_predict_async` invocation. Additionally, even the synchronous `fit_predict` path never forwarded `coverage` to `executor.predict()`, so prediction intervals were silently dropped everywhere.

### `src/sktime_mcp/tools/fit_predict.py`
- Added `coverage: Optional[Any] = None` to `fit_predict_tool` and forwarded it to `executor.fit_predict()`
- - Added `coverage: Optional[Any] = None` to `fit_predict_async_tool` and forwarded it to `executor.fit_predict_async()`
### `src/sktime_mcp/runtime/executor.py`
- Added `coverage` parameter to `executor.predict()` - calls `predict_interval()` when coverage is set and flattens the MultiIndex result to readable string keys
- - Added `coverage` parameter to `executor.fit_predict()` - passes it to `predict()`
- - Added `coverage` parameter to `executor.fit_predict_async()` - forwards it to the background predict step
## Behavior After Fix

```python
# Async fit-predict with prediction intervals now works correctly
fit_predict_async(estimator_handle="est_123", dataset="airline", coverage=0.9)
# -> job result includes both predictions AND prediction_intervals
```

## Testing

- `ruff check` passes on all modified files
- - `make check` passes locally
- - Manually verified the call chain by tracing coverage through all layers


Shashankss1205:
The tools this PR patches (fit_predict, fit_predict_async) were removed in the consolidation (#460); the current predict tool returns prediction intervals via mode='predict_interval' and coverage, and runs in the background with run_async=True. The branch also carries unrelated CI/docs changes. Closing; #176 is closed as superseded.

Issue #148, opened 2026-04-12
[BUG] CI pipeline fails on main lint errors, formatting issues, broken test, and Python 3.9 incompatibility


Abhishek9639:
**Describe the bug**
The CI pipeline is currently failing on all 6 checks (Lint, Test Python 3.9/3.10/3.11/3.12, Docs Build) even on clean branches based off main. There are four separate pre-existing issues causing this:
1. **Lint errors** - `ruff check` fails due to unsorted imports in `server.py` and `test_evaluate.py`, a missing trailing newline in `instantiate.py`, and a whitespace/style issue in `test_evaluate.py`
2. **Formatting issues** - `ruff format --check` fails because `evaluate.py`, `instantiate.py`, and `test_evaluate.py` need reformatting
3. **Broken test** - `test_evaluate_estimator_tool` in `test_evaluate.py` asserts `result["cv_folds_run"] == 2` but the actual value returned is `4`
4. **Python 3.9 incompatibility** - `pyproject.toml` sets `requires-python = ">=3.9"` but the `mcp` dependency requires Python 3.10+, so `pip install` fails on 3.9

**To Reproduce**

```
# Lint check
ruff check .
# 5 errors in server.py, instantiate.py, test_evaluate.py
# Format check
ruff format --check .
# 3 files need reformatting: evaluate.py, instantiate.py, test_evaluate.py
# Test
pytest tests/test_evaluate.py -v
# FAILED assert 4 == 2
# Python 3.9 install
# pip install -e ".[dev]"  on Python 3.9
# ERROR: No matching distribution found for mcp>=1.0.0

```
**Expected behavior**
All CI checks should pass on the main branch. Lint and formatting should be clean, the test assertion should match the actual behavior, and the minimum Python version should be compatible with all dependencies.

**Additional context**
I noticed this while working on #136 my PR #147 has all 6 checks failing but none of the failures are related to my changes. They all reproduce on a clean checkout of main.

Happy to submit a fix for these if that would be helpful.

**Versions**
Python: 3.13.11 (also tested with 3.9 for the mcp issue)
sktime: 1.0.0 (from pyproject.toml)
OS: macOS / Ubuntu (GitHub Actions)

Abhishek9639:
Greetings @Shashankss1205,
I'd like to work on this issue and fix the calculation logic.
I will be opening a PR shortly.
Thanks

direkkakkar319-ops:
[...]

PR #179, opened 2026-04-13
[ENH] Unify fit_predict and fit_predict_with_data tools (#4)

### LINKED ISSUE #4: [ENH] Unify `fit_predict` and `fit_predict_with_data` tools
**Description**
The MCP server currently exposes two separate tools for essentially the same logic: `fit_predict` (restricted to demo datasets) and `fit_predict_with_data` (for custom loaded data handles). This forces the LLM to choose between two different tools based on the data source. We should unify these into a single, smarter `fit_predict` tool that can handle both.

**Suggested Approach**
- Update the `fit_predict` tool in `src/sktime_mcp/tools/fit_predict.py` to accept either `dataset` (demo string) or `data_handle` (string ID).
- Modify the `Executor.fit_predict` implementation to branch based on which input is provided.
- Update the tool schema in `server.py` and mark `fit_predict_with_data` as redundant/deprecated.

**Ponder Upon**
Is there a better way than the one suggested above? Open to discussion in this thread.

Abhishek9639:
### Reference Issues/PRs

Fixes #4

This PR completes the unification of `fit_predict` and `fit_predict_with_data` tools.

The `fit_predict` tool already accepts both `dataset` (demo string) and `data_handle` (custom data ID), so the separate `fit_predict_with_data` tool is now redundant. This PR marks it as deprecated at all three layers.

- **`data_tools.py`** - `fit_predict_with_data_tool()` now logs a deprecation warning and delegates to `fit_predict_tool(data_handle=...)` instead of calling the executor directly.
- **`executor.py`** - `Executor.fit_predict_with_data()` now logs a deprecation warning and delegates to `self.fit_predict(data_handle=...)`.
- **`server.py`** - Removed the direct import of `fit_predict_with_data_tool`. The backward-compat dispatcher handler now routes through `fit_predict_tool` directly.

Backward compatibility is fully maintained old `fit_predict_with_data` calls still work, they just emit a warning and delegate to the unified path.

No.

- The deprecation approach delegating to the unified `fit_predict_tool` rather than deleting the old code outright, to avoid breaking existing LLM clients.
- Backward compatibility of the dispatcher in `server.py`.

### Any other comments?

The 1 test failure (`test_evaluate.py`) is a pre-existing issue on `main` tracked by #148 and is unrelated to this change.

##### For all contributions

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -831,6 +831,11 @@ class Executor:
         """
         Fit and predict using a data handle.
 
+        .. deprecated::
+            Use ``fit_predict(handle_id, dataset="", horizon=horizon,
+            data_handle=data_handle)`` instead. This method will be removed
+            in a future release.
+
         Args:
             estimator_handle: Estimator handle from instantiate_estimator
             data_handle: Data handle from load_data_source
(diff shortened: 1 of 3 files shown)


Abhishek9639:
Greetings @Shashankss1205, 
This PR is ready for review.

The `fit_predict` tool already supports both `dataset` and `data_handle` inputs, so `fit_predict_with_data` was redundant. I've marked it as deprecated across all layers (tool function, executor method, and server dispatcher) while keeping backward compatibility intact.

Let me know if you'd like any changes. 
Thanks

Shashankss1205:
I am still unsure why are you not removing the "fit_predict_with_data"  entirely rather just doing a cross fn to tool relation
call_tool should never ever think of "call_tool" independently, I am making a fix for this from my end because it is blocking many more future PRs. Thanks for your time.

Issue #194, opened 2026-04-13
[ENH] Add SKTIME_MCP_LOG_LEVEL env var - logging hardcoded to DEBUG causes stderr noise in MCP clients


being-iota:
## Summary

`server.py` hardcodes `logging.basicConfig(level=logging.DEBUG)`, causing every tool call - including full JSON argument and result payloads - to be written to stderr unconditionally.

In MCP environments like Claude Desktop, stderr is captured and surfaced to the user. DEBUG-level output buries real errors and exposes potentially sensitive data (file paths, data handles) with no way to reduce verbosity.

> In MCP systems, stderr is effectively part of the user interface, not just a debugging channel.

## Evidence

Current hardcoded implementation in `server.py`:

```python
logging.basicConfig(
    level=logging.DEBUG,  # no env var, no override possible
    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
    handlers=[logging.StreamHandler()],
)
```

The same file already uses env vars for configuration:

```python
JOB_MAX_AGE_HOURS = int(os.environ.get("SKTIME_MCP_JOB_MAX_AGE_HOURS", "24"))
JOB_CLEANUP_INTERVAL_SECS = int(os.environ.get("SKTIME_MCP_JOB_CLEANUP_INTERVAL", "3600"))
```

Logging does not follow this established pattern.

## Impact on MCP Workflows

- Every tool call logs full JSON payloads to stderr
- Real errors are buried in debug noise in Claude Desktop
- Sensitive information (paths, data handles) always logged
- No way to reduce verbosity without modifying source code

## Suggested Fix

```python
_LOG_LEVEL = os.environ.get("SKTIME_MCP_LOG_LEVEL", "WARNING").upper()
logging.basicConfig(
    level=getattr(logging, _LOG_LEVEL, logging.WARNING),
    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
    handlers=[logging.StreamHandler()],
)
```

## Expected Behaviour

- Default level becomes `WARNING` - clean output in MCP clients
- Developers restore verbose logs via `SKTIME_MCP_LOG_LEVEL=DEBUG`
- Follows the existing env-var pattern already in this file

Happy to submit a PR immediately.

Shashankss1205:
Already fixed in the latest version.

PR #198, opened 2026-04-14
fix: sanitize_for_json now natively supports NumPy and Pandas types

### LINKED ISSUE #197: [BUG] sanitize_for_json crashes when returning NumPy or Pandas types
## Describe the bug
The `sanitize_for_json` utility function in `src/sktime_mcp/server.py` does not safely serialize NumPy or Pandas native types. If an MCP tool returns NumPy integers (e.g., `np.int64`), floating numbers (`np.float64`), or any Pandas Timestamps (`pd.Timestamp`), it falls back to the default handler and crashes the JSON serializer in the MCP server stdio layer.

## Expected behavior
The JSON sanitizer should seamlessly downcast common data science types into standard Python primitives (e.g., `np.int64` -> `int` and `pd.Timestamp` -> `str`).

## Actual behavior
Any raw tool output containing these types triggers a server crash that interrupts LLM processing:

```python
TypeError: Object of type int64 is not JSON serializable

### PR DESCRIPTION
**Pull request by SakshiKasat18**
**Fixes #197**

Adds numpy and pandas type parsing to avoid JSON serialization TypeErrors. Includes comprehensive test suite.

### Reference Issues/PRs

Fixes #197 

This PR fixes a bug where `sanitize_for_json` in `server.py` crashes the MCP server when attempting to serialize Data Science types back into JSON for the LLM. 

Because `sktime` pipelines and evaluation tools often return `np.int64`, `np.float64`, `pd.Timestamp`, or `pd.NA` natively inside dictionary results (e.g. `eval_results.to_dict()`), the `json.dumps()` in the `stdio_server` previously threw a `TypeError: Object of type int64 is not JSON serializable`.

This adds safe, recursive type coercion for:
* `np.integer` -> `int`
* `np.floating` -> `float`
* `np.ndarray` -> `list`
* `np.bool_` -> `bool`
* `pd.Timestamp` -> ISO `str`
* `pd.NA` / `pd.NaT` -> `None`
* `pd.Series` / `pd.DataFrame` -> `list` / `dict`

A comprehensive 27-check test suite was added in `test_sanitize.py` to ensure complete coverage.

No, the implementation uses optional soft imports (`try/except ImportError`) for both `numpy` and `pandas` to ensure it doesn't enforce strict dependencies if they aren't globally available.

Please verify `test_sanitize.py` covers the expected breadth of objects that the MCP tools generally output.

### PR checklist

**For all contributions**
- [x] I've added myself to the [list of contributors](https://github.com/sktime/sktime/blob/main/CONTRIBUTORS.md).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/sktime/sktime/blob/main/.github/CODEOWNERS) to receive notifications about future changes to these files.

**For new estimators**

### DIFF
--- a/src/sktime_mcp/server.py
+++ b/src/sktime_mcp/server.py
@@ -11,6 +11,18 @@ import logging
 import os
 from typing import Any
 
+try:
+    import numpy as np
+    _NUMPY_AVAILABLE = True
+except ImportError:
+    _NUMPY_AVAILABLE = False
+
+try:
+    import pandas as pd
+    _PANDAS_AVAILABLE = True
+except ImportError:
+    _PANDAS_AVAILABLE = False
+
 from mcp.server import Server
 from mcp.server.stdio import stdio_server
 from mcp.types import TextContent, Tool
@@ -56,8 +68,9 @@ JOB_MAX_AGE_HOURS = int(os.environ.get("SKTIME_MCP_JOB_MAX_AGE_HOURS", "24"))
 JOB_CLEANUP_INTERVAL_SECS = int(os.environ.get("SKTIME_MCP_JOB_CLEANUP_INTERVAL", "3600"))
 
 # Configure logging to stderr with detailed format
+_LOG_LEVEL = os.environ.get("SKTIME_MCP_LOG_LEVEL", "WARNING").upper()
 logging.basicConfig(
-    level=logging.DEBUG,
+    level=getattr(logging, _LOG_LEVEL, logging.WARNING),
     format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
     handlers=[logging.StreamHandler()],
 )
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by SarthakB11 — COMMENTED** [review]
Thanks for tackling this — the overall structure is right and this clearly fixes the common numpy/pandas crash cases.

A couple of things worth addressing before merge:

**1. NaN and Inf are still not JSON-serializable**

The current fix does `return float(obj)` for `np.floating`, but `float('nan')` and `float('inf')` are not valid JSON. `json.dumps({'x': float('nan')})` still raises `ValueError: Out of range float values are not JSON compliant`. A quick repro:

```python
import numpy as np, json
result = sanitize_for_json({'metric': np.float64('nan')})
json.dumps(result)  # ValueError: Out of range float values are not JSON compliant
```

Suggested fix: map NaN/inf to `None` (JSON `null`):
```python
if isinstance(obj, np.floating):
    f = float(obj)
    if not (f == f) or f == float('inf') or f == float('-inf'):
        return None
    return f
```

**2. `pd.NA` check is duplicated**

Lines 104–109 have the `pd.NA` check twice — looks like a copy-paste artefact.

**3. Scope: unrelated `SKTIME_MCP_LOG_LEVEL` change**

The log level env var wiring is a useful fix, but it belongs to a separate issue (#194). Bundling it here means this PR can't be merged independently if the log level change needs more discussion. Worth splitting out.

**4. Test file location**

`test_sanitize.py` is placed at the repo root rather than in `tests/` — the CI test runner likely won't pick it up there.

---

For what it's worth, I ran into the NaN/inf issue myself and opened a follow-up in #275 that addresses points 1 and 2. Happy to close that if you'd prefer to fold the fixes here instead.

Issue #170, opened 2026-04-13
[BUG] CI workflow broken. mkdocs build fails after Sphinx migration, Python 3.9 still in test matrix


areychana:
**Describe the bug**
Two issues in `.github/workflows/ci.yml` are causing every CI run to fail (0/6 checks):

1. The Docs Build job still runs `mkdocs build` (line 70) but the project was recently migrated from mkdocs to Sphinx (via commits 9f032f8, c85f7b5). This causes Docs Build to fail on every commit and PR.

2. The test matrix still includes Python 3.9 (line 37) but `pyproject.toml` was updated to require `>=3.10` via #164. Installing dependencies on 3.9 fails because `mcp>=1.0.0` does not support Python 3.9.

**To Reproduce**
Any push or PR to main triggers CI and all runs currently show 0/6 checks passing. 

**Expected behavior**
- Docs Build should use `sphinx-build docs/source docs/_build/html` to match the new
Sphinx setup (config at `docs/source/conf.py` per `.readthedocs.yaml`)
- Python 3.9 should be removed from the test matrix to match `requires-python = ">=3.10"`

**Additional context**
Both fixes are isolated to `.github/workflows/ci.yml`, no other files need changing.

**Versions**
N/A. this is a CI configuration bug

PR #206, opened 2026-04-15
fix: propagate coverage param through predict/fit_predict/fit_predict…

### LINKED ISSUE #205: [BUG] fit_predict_async silently drops coverage parameter — prediction intervals never returned for async jobs
## Summary
`executor.fit_predict_async()` has no `coverage` parameter in its signature (executor.py:249), but `fit_predict_async_tool` tries to pass `coverage=coverage` to it (fit_predict.py:183). The argument is silently dropped, and `self.predict()` is called without `coverage` (executor.py:343), so prediction intervals are **never returned** for async jobs even when explicitly requested.

## Root Cause
`fit_predict_async` signature (executor.py:249):
```python
async def fit_predict_async(self, handle_id, dataset, horizon=12, job_id=None):
    # no coverage param!

rupeshca007:
## Summary

Fixes #205

`executor.predict()` and `executor.fit_predict()` were missing the `coverage` parameter entirely, and `executor.fit_predict_async()` had no `coverage` in its signature. Calling `fit_predict_async_tool(coverage=0.9)` silently dropped the argument — prediction intervals were never computed for async jobs.

## Changes (single file: `src/sktime_mcp/runtime/executor.py`)

1. Added `coverage` param to `executor.predict()` with prediction interval computation
2. Added `coverage` param to `executor.fit_predict()` and forwarded it to `self.predict()`
3. Added `coverage` param to `executor.fit_predict_async()` and forwarded it to `self.predict()`

## Testing

- `test_async_fit_predict` now passes (was failing with `TypeError: Executor.predict() got an unexpected keyword argument 'coverage'`)
- All 70 other tests continue to pass

## Type of Change
- [x] Bug fix (non-breaking)

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -148,8 +148,9 @@ class Executor:
         handle_id: str,
         fh: Optional[Union[int, list[int]]] = None,
         X: Optional[Any] = None,
+        coverage: Optional[Union[float, list[float]]] = None,
     ) -> dict[str, Any]:
-        """Generate predictions."""
+        """Generate predictions, and optionally prediction intervals."""
         try:
             instance = self._handle_manager.get_instance(handle_id)
         except KeyError:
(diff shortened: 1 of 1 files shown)


rupeshca007:
"CI failures are pre-existing on main — tracked in #170/#171. My change only touches executor.py."

Shashankss1205:
Since then the tool surface was consolidated: fit_predict/fit_predict_async are gone, replaced by separate fit and predict tools with a run_async flag, and predict already accepts mode='predict_interval' with coverage (and predict_quantiles with alpha). The code this PR patches no longer exists. Closing; #205 is closed as well.

PR #208, opened 2026-04-15
[BUG] fix: correct initial_window formula so cv_folds is respected in evaluate_estimator_tool

### LINKED ISSUE #207: [BUG] evaluate_estimator_tool ignores cv_folds — wrong initial_window formula returns more folds than requested
## Describe the bug

`evaluate_estimator_tool` does not respect the `cv_folds` parameter. Requesting `cv_folds=2` returns 4 folds instead of 2. This also causes `test_evaluate_estimator_tool` to fail in CI with `assert 4 == 2`.

## Root Cause

In `src/sktime_mcp/tools/evaluate.py`, line 51:

```python
initial_window = max(int(n * 0.5), n - cv_folds * 2)

rupeshca007:
## Summary

Fixes #207

The `initial_window` formula in `evaluate_estimator_tool` was incorrect, causing more cross-validation folds to be generated than the user requested. This also caused `test_evaluate_estimator_tool` to fail in CI with `assert 4 == 2`.

## Root Cause

**Before (wrong):**
```python
initial_window = max(int(n * 0.5), n - cv_folds * 2)


rupeshca007:
The 3 CI failures (Python 3.9, Lint, Docs Build) are pre-existing on main and tracked in issue #170. They are not caused by this change. Python 3.10, 3.11, and 3.12 all pass. The fix can be verified cleanly once PR #171 merges.

PR #216, opened 2026-04-15
fix(validator): disallow forecaster-to-forecaster chains in pipeline validation

### LINKED ISSUE #201: [BUG] validate_pipeline allows forecaster → forecaster chains in pipeline validation
## Summary
`validate_pipeline` currently treats forecaster → forecaster compositions as valid in a pipeline context, which is not correct for linear pipelines.

## Description
While testing pipeline validation, I observed that chaining two forecasters is marked as valid. This contradicts expected pipeline behavior and also conflicts with the example provided in `examples/01_forecasting_workflow.py`, where this case is explicitly labeled as invalid.

## Steps to Reproduce
Run the following:

```python
from sktime_mcp.composition.validator import get_composition_validator

validator = get_composition_validator()

result = validator.validate_pipeline(
    ["NaiveForecaster", "ExponentialSmoothing"]
)

print("valid =", result.valid)
print("errors =", result.errors)
print("warnings =", result.warnings)
print("suggestions =", result.suggestions)
```

## Observed Behavior
```
valid = True
errors = []
warnings = []
suggestions = []
```

## Expected Behavior
The pipeline:

```python
["NaiveForecaster", "ExponentialSmoothing"]
```

should be considered invalid in a linear pipeline context, with an appropriate error message.

This expectation is also reflected in:

- `examples/01_forecasting_workflow.py` (where this case is labeled as invalid)

## Analysis
The validation logic appears to allow forecasting → forecasting transitions without distinguishing between pipeline and ensemble contexts. As a result, a rule intended for a different composition scenario is being applied during pipeline validation.

## Impact
This can affect agentic workflows where LLMs rely on `validate_pipeline` to construct valid pipelines. Incorrect validation may lead to:

- Invalid pipeline construction
- Downstream runtime issues
- Reduced reliability of automated pipeline generation

## Contribution
I plan to raise a PR once I have a fix ready. In the meantime, I would appreciate confirmation on whether this behavior is unintended.

ParminderSinghGithub:
Closes #201 

This PR attempts to resolve the issue by correcting pipeline validation behavior. Happy to make any changes if needed.

## Summary
This PR fixes an issue where `validate_pipeline` incorrectly allowed chaining one forecaster after another in a linear pipeline.

## What was wrong
The validation logic treated forecasting → forecasting as valid because it matched a rule intended for ensemble usage. However, this rule should not apply to sequential pipelines.

## What was changed
- Added a check in `validate_pipeline` to prevent forecaster → forecaster chaining in linear pipelines
- Kept the existing error message unchanged
- Left ensemble-related rules untouched

## Tests
- Added a regression test to ensure forecaster chaining is correctly rejected
- Added a test to confirm valid transformer → forecaster pipelines still work as expected

## Why this matters
`validate_pipeline` is used in agentic workflows to construct valid pipelines. Allowing invalid combinations can lead to incorrect pipeline construction and runtime issues. This change ensures such cases are properly caught.

### DIFF
--- a/src/sktime_mcp/composition/validator.py
+++ b/src/sktime_mcp/composition/validator.py
@@ -213,6 +213,16 @@ class CompositionValidator:
             current_name, current_node = nodes[i]
             next_name, next_node = nodes[i + 1]
 
+            # In linear pipelines, forecaster -> forecaster is invalid.
+            # The forecasting->forecasting rule is for ensemble composition,
+            # not sequential pipeline chaining.
+            if current_node.task == next_node.task == "forecasting":
+                errors.append(
+                    f"Cannot chain forecasters '{current_node.name}' → '{next_node.name}' directly. "
+                    "Use an ensemble or multiplexer instead."
+                )
+                continue
+
             # Check if this composition is valid
             valid_pair, pair_errors, pair_warnings = self._check_pair_compatibility(
                 current_node, next_node
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -123,6 +123,26 @@ class TestCompositionValidator:
 
         assert not result.valid
 
+    def test_forecaster_chain_invalid(self):
+        """Forecaster -> forecaster should be invalid in linear pipeline validation."""
+        from sktime_mcp.composition.validator import CompositionValidator
+
+        validator = CompositionValidator()
+        result = validator.validate_pipeline(["NaiveForecaster", "ExponentialSmoothing"])
+
+        assert not result.valid
+        assert any("Cannot chain forecasters" in err for err in result.errors)
+
+    def test_transformer_to_forecaster_valid(self):
+        """Transformer -> forecaster remains valid."""
+        from sktime_mcp.composition.validator import CompositionValidator
+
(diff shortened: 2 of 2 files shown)


Saloni-0465 (commented):
This looks like a solid fix for incorrect forecaster chaining 👍

Would it make sense to extend this further by distinguishing between “valid” and “buildable” pipelines (e.g., via a `buildable` flag in the validator)? That could help prevent similar runtime issues more generally.

ParminderSinghGithub:
Thanks for the kind words, and great point — I think you're right that this is worth pursuing properly.

For now, I've kept this PR scoped to fixing the forecaster chaining behavior in linear pipelines, since introducing a `buildable` signal would likely touch the validator contract and how results are consumed downstream — changes that deserve their own focused review.

That said, I'm genuinely interested in taking this further. Once this is merged, I'll open a follow-up issue and then a PR with a concrete proposal, starting with where a `buildable` flag could fit into the current validator output without breaking existing consumers. Happy to loop you in when the time comes!

Shashankss1205 (approved):

PR #227, opened 2026-04-17
fix: exclude loaders with required args from demo dataset registry

### LINKED ISSUE #226: [BUG] list_datasets includes loaders with required arguments, causing runtime failures
# [BUG] list_datasets includes loaders with required arguments, causing runtime failures
## Summary

`_discover_demo_datasets` picks up every `load_*` function from `sktime.datasets` by name,
including ones that require mandatory arguments like a file path or dataset name.
These show up in `list_datasets()` as valid options, but calling any of them fails immediately
with a raw `TypeError` from Python. 9 out of 33 listed datasets (27%) are broken this way.

## Steps to Reproduce

```python
import sys
sys.path.insert(0, "src")

executor = get_executor()

# These all appear in list_datasets() output:
for name in ["UCR_UEA_dataset", "from_tsfile", "fpp3", "forecastingdata"]:
    result = executor.load_dataset(name)
    print(name, "->", result["error"])
```

Output:
```
UCR_UEA_dataset -> load_UCR_UEA_dataset() missing 1 required positional argument: 'name'
from_tsfile     -> load_from_tsfile() missing 1 required positional argument: 'full_file_path_and_name'
fpp3            -> load_fpp3() missing 1 required positional argument: 'dataset'
forecastingdata -> load_forecastingdata() missing 1 required positional argument: 'name'
```

## Expected vs Actual

| | Expected | Actual |
|---|---|---|
| Datasets in `list_datasets()` | Only zero-argument loaders | All 33 `load_*` functions including 9 that need arguments |
| Error when loading a broken dataset | A clear message explaining the dataset needs a file path / is not a demo dataset | Raw `TypeError` leaking internal function signature |

## What's Happening

[`executor.py`](https://github.com/sktime/sktime-mcp/blob/main/src/sktime_mcp/runtime/executor.py) discovers datasets like this:

```python
return {
    name.removeprefix("load_"): f"sktime.datasets.{name}"
    for name, obj in inspect.getmembers(_ds_module, inspect.isfunction)
    if name.startswith("load_")
}
```

This includes loaders that aren't demo datasets at all, they're file/path utilities that
happen to start with `load_`. When called with no arguments they raise a `TypeError`,
which bubbles up as the error message.

The full list of broken entries currently included:

| Name in registry | Missing argument |
|---|---|
| `UCR_UEA_dataset` | `name` |
| `forecastingdata` | `name` |
| `fpp3` | `dataset` |
| `from_arff_to_dataframe` | `full_file_path_and_name` |
| `from_long_to_dataframe` | `full_file_path_and_name` |
| `from_tsfile` | `full_file_path_and_name` |
| `from_tsfile_to_dataframe` | `full_file_path_and_name` |
| `from_ucr_tsv_to_dataframe` | `full_file_path_and_name` |
| `tsf_to_dataframe` | `full_file_path_and_name` |

## Suggested Fix

Filter out any loader that has required positional arguments:

```python
import inspect

return {
    name.removeprefix("load_"): f"sktime.datasets.{name}"
    for name, obj in inspect.getmembers(_ds_module, inspect.isfunction)
    if name.startswith("load_")
    and all(
        p.default is not inspect.Parameter.empty
        for p in inspect.signature(obj).parameters.values()
        if p.kind not in (p.VAR_POSITIONAL, p.VAR_KEYWORD)
    )
}
```

This would bring the registry down to only the genuinely zero-argument demo loaders,
which is what the tool is meant to expose.

## Environment

- Python: `>=3.10`

Happy to submit a PR for this since the fix is a one-liner in _discover_demo_datasets. Let me know if you'd prefer a different approach.

Om-Dhondge:
## Summary
- Filter `_discover_demo_datasets` to only include zero-argument `load_*` functions
- Reduces the exposed dataset list from 33 to 24 (removes 9 file/path utilities that were never valid demo datasets)

## Why
`_discover_demo_datasets` picked up every `load_*` function from `sktime.datasets`, including
ones like `load_from_tsfile(full_file_path_and_name)` that require mandatory arguments. These
appeared in `list_datasets()` as valid options but always failed with a raw `TypeError` when
called. The fix uses `inspect.signature` to include only zero-argument loaders.

Closes #226 

## Test plan
- [ ] `list_datasets()` no longer includes `from_tsfile`, `UCR_UEA_dataset`, `fpp3`, `forecastingdata`, and the other 5 file-path loaders
- [ ] All previously working datasets (e.g. `airline`, `lynx`) still load successfully

Looking forward to your approval, thanks ! 
@Shashankss1205

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -26,7 +26,7 @@ logger = logging.getLogger(__name__)
 # load_* function in sktime.datasets to the MCP server.
 def _discover_demo_datasets() -> dict:
     """Return a mapping of dataset name -> dotted module path for every
-    ``load_*`` function exported by ``sktime.datasets``."""
+    zero-argument ``load_*`` function exported by ``sktime.datasets``."""
     try:
         import sktime.datasets as _ds_module
 
@@ -34,6 +34,11 @@ def _discover_demo_datasets() -> dict:
             name.removeprefix("load_"): f"sktime.datasets.{name}"
             for name, obj in inspect.getmembers(_ds_module, inspect.isfunction)
             if name.startswith("load_")
+            and all(
+                p.default is not inspect.Parameter.empty
+                for p in inspect.signature(obj).parameters.values()
+                if p.kind not in (p.VAR_POSITIONAL, p.VAR_KEYWORD)
+            )
         }
     except Exception:  # pragma: no cover
         return {}  # fallback: empty dict if sktime not installed


Shashankss1205:
Verified: still applies to current main; the catalogue drops from 33 to 24 entries, removing exactly the loaders that need positional arguments while keeping airline, lynx, shampoo_sales, longley. Full suite green. Merging; #235 is a duplicate. Network-only loaders and task mis-bucketing are tracked in the follow-up issue.

Shashankss1205:
Landed on main as f6fb797 (squash-merged locally because the org policy rejects the merge API for this token). Thanks!

PR #254, opened 2026-04-20
Memory leak load data source

### LINKED ISSUE #218: [BUG] Memory leak in load_data_source: original data handle never freed
# [BUG] Memory leak in load_data_source: original data handle never freed

## Summary

Every call to `load_data_source` with auto-formatting enabled (the default) leaves an unreachable data handle in memory. The original handle is created, passed to `format_data_handle`, and then abandoned, and only the formatted handle is returned to the caller. Over a session with repeated data loads, this could accumulate DataFrame-sized objects without a way to release them.

## Steps to Reproduce

```python
import sys
sys.path.insert(0, "src")

import pandas as pd
from sktime.datasets import load_airline

executor = get_executor()

y = load_airline()
df = pd.DataFrame({"date": y.index.to_timestamp(), "passengers": y.values})

result = executor.load_data_source({
    "type": "pandas",
    "data": df,
    "time_column": "date",
    "target_column": "passengers",
})

print("Returned handle:", result["data_handle"])
print("Orphaned handle:", result["original_handle"])
print("Total handles in memory:", len(executor._data_handles))
# Total handles in memory: 2  ← expected 1
```

## Expected vs Actual Behaviour

| | Expected | Actual |
|---|---|---|
| Handles in memory after one load | 1 | 2 |
| `original_handle` in response | not present / same as `data_handle` | a separate unreachable handle |
| Memory freed when session ends | yes | only via GC at process exit |

## What's Happening

In [`executor.py`](https://github.com/sktime/sktime-mcp/blob/main/src/sktime_mcp/runtime/executor.py) the `load_data_source` method:

1. Creates `data_handle` and stores the raw DataFrame under it in `_data_handles` (~line 533).
2. Calls `format_data_handle(data_handle)`, which creates a second handle `new_handle` and stores the formatted copy (~line 799).
3. Returns `new_handle` to the caller, but  it **never deletes `data_handle`**.

```python
# ~line 533
data_handle = f"data_{uuid.uuid4().hex[:8]}"
self._data_handles[data_handle] = { "y": y, "X": X, ... }  # stored

# ~line 547
format_result = self.format_data_handle(data_handle, ...)
if format_result["success"]:
    return {
        "data_handle": format_result["data_handle"],  # new handle returned
        "original_handle": data_handle,               # original exposed but never cleaned up
        ...
    }
# data_handle stays in _data_handles forever
```

The caller receives `format_result["data_handle"]` and has no reason to release `data_handle`. Even though `original_handle` is present in the response, there's no prompt to release it, and the MCP tool schema doesn't document this.

## Impact

Each `load_data_source` call (auto-format is on by default) leaks one full copy of the loaded DataFrame. In an agentic session that loads data repeatedly with different datasets, different data slices, and retries this could easily get compounded quickly. The existing issue #191 tracks the broader lack of eviction; however, this is a specific structural source of that accumulation.

## Suggested Fix

Delete the original handle immediately after a successful format:

```python
if format_result["success"]:
    del self._data_handles[data_handle]  # release pre-format copy
    return {
        "data_handle": format_result["data_handle"],
        ...
    }
```

If keeping the original for rollback/debugging is intentional, that should be documented, and a `release_data_handle` call should be explicitly recommended in the response.

## Environment

- Python: `>=3.10`

Nischal1425:
#### Reference Issues/PRs
Fixes #218

Each call to `load_data_source` with auto-formatting enabled (the default)
leaked one full DataFrame-sized object in memory per call.

**Root cause:** When auto-formatting succeeds, a new formatted handle is
created and returned — but the original raw handle was never removed from
`self._data_handles`, leaving an unreachable DataFrame accumulating in
memory for the entire session.

**Fix:** In `executor.py`, added `del self._data_handles[data_handle]`
immediately after successful formatting. Also removed the now-invalid
`"original_handle"` field from the response, since that handle no longer
exists after deletion.

No.

- The single `del self._data_handles[data_handle]` line in `executor.py`
  (~line 552) — confirm this is the right place and that no other code path
  holds a reference to the raw handle after formatting succeeds.
- The three new tests in `TestMemoryLeakFix` in `tests/test_core.py`.

The fix was verified with three tests:
1. After one load, exactly 1 handle exists (not 2).
2. `"original_handle"` is no longer present in the response.
3. Loading N times yields exactly N handles, not 2×N.

All 38 existing + new tests pass locally.


Nischal1425:
@felipeangelimvieira resolved the merge conflict in tests/test_core.py (kept both TestMemoryLeakFix and TestSearchEstimatorsLimit). Ready for review.

Shashankss1205 (approved):

PR #255, opened 2026-04-20
fix: use get_params() for deep parameter key validation in instantiate_estimator

### LINKED ISSUE #219: [ENH] Deep parameter validation for instantiate_estimator (Follow-up to #3)
Hi @Shashankss1205! 

I was looking through the recently closed issues and noticed your "Ponder Upon" question in #3: *"Can we check if all the keys in the dict belong to the estimator object or not?"*

PR #13 cleanly handled the base `isinstance` type check, but the deeper parameter key validation is still open. Currently, if a user passes invalid keys to the estimator, it silently fails or throws confusing internal errors, which isn't ideal (especially for LLM workflows). 

I've drafted a solution that validates the `params` keys against the estimator's actual `get_params()` and returns a clear, actionable `ValueError` listing the valid parameters. 

Would you like me to open a PR for this to fully resolve the original idea?

Nischal1425:
#### Reference Issues/PRs
Fixes #219. See also #3 (original type validation PR).

Previously, `instantiate_estimator_tool` validated param *types* but not
param *keys*. Passing a typo'd key like `{"window_leangth": 5}` would
silently create a broken estimator with no feedback — especially problematic
in LLM workflows where the model can't recover from silent failures.

A partial fix existed (lines 82–97 of `instantiate.py`) that warned using
registry metadata derived from `__init__` signatures, but:
- It only warned, never errored
- Registry metadata can be incomplete or stale

**Fix:** Replaced the warning-only approach with a proper error using the
estimator's actual `get_params(deep=False)` output from a default instance.
This is the canonical sktime API for discovering valid parameters.

Fallback: if the class can't be instantiated with defaults, we fall back to
registry hyperparameter keys (existing behaviour) to avoid false negatives.

Unknown keys now return `success=False` with a message listing all valid
parameter names, e.g.:
Unknown parameter(s) for NaiveForecaster: ['window_leangth'].
Valid parameters are: ['sp', 'strategy', 'window_length']

No.

- Lines 82–111 of `instantiate.py` — the `get_params()` deep validation block.
- The fallback logic when `cls()` instantiation raises an exception.
- Whether changing from warning → error is the right severity, or if a
  stricter/looser approach is preferred.

Updated `tests/test_param_validation.py`:
- `test_unknown_key_produces_warning` → replaced with `test_unknown_key_returns_error`
- Added `test_valid_key_passes`, `test_typo_in_key_returns_error_with_valid_list`,
  `test_no_estimator_name_skips_deep_validation`
- Added `test_unknown_key_returns_error_with_valid_list` to `TestInstantiateEstimatorValidation`

All 38 tests pass locally.

### DIFF
--- a/src/sktime_mcp/tools/instantiate.py
+++ b/src/sktime_mcp/tools/instantiate.py
@@ -79,22 +79,36 @@ def _validate_params(
                 "warnings": warnings,
             }
 
-    # check if keys match known hyperparameters (warn, don't error)
+    # deep validation: check keys against estimator's actual get_params()
     if estimator_name and params:
         registry = get_registry()
         node = registry.get_estimator_by_name(estimator_name)
 
-        if node is not None and node.hyperparameters:
-            known_keys = set(node.hyperparameters.keys())
-            provided_keys = set(params.keys())
-            unknown_keys = provided_keys - known_keys
+        if node is not None:
+            valid_keys = None
 
-            if unknown_keys:
-                warnings.append(
-                    f"Unknown parameter(s) for {estimator_name}: "
-                    f"{sorted(unknown_keys)}. "
-                    f"Known parameters: {sorted(known_keys)}"
-                )
+            # prefer get_params() from a default instance — more reliable than
+            # registry metadata which is derived from __init__ signatures only
+            try:
+                cls = node.estimator_class
+                valid_keys = set(cls().get_params(deep=False).keys())
+            except Exception:
+                # fall back to registry hyperparameter keys if instantiation fails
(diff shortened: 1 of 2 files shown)


Nischal1425:
@felipeangelimvieira this PR is ready for review when you get a chance.

Shashankss1205:
The instantiate_estimator tool and its _validate_params helper were removed in the refactor; instantiate now builds objects from a craft() spec and an unknown parameter fails at once with 'unexpected keyword argument', which resolves the silent-failure concern in #219. Closing as obsolete (same as #256).

PR #256, opened 2026-04-20
[ENH] Validate estimator param keys in instantiate_estimator

Rahul-2k4:
## Summary
This addresses #219 by validating `params` keys before estimator instantiation.

Previously, unknown keys made it to constructor execution and failed later with a generic unexpected-keyword error. With this change, the tool fails early with a clear validation message that lists:
- unknown keys
- valid keys for the target estimator

## Changes
- Updated `_validate_params` in `src/sktime_mcp/tools/instantiate.py` to reject unknown keys.
- Kept the change scoped to validation behavior; no broad API redesign.
- Expanded tests in `tests/test_param_validation.py`.

### RED
Failing tests before implementation:
- `TestValidateParams::test_unknown_key_returns_error`
- `TestInstantiateEstimatorValidation::test_unknown_key_returns_error`

### GREEN
Command:
- `TMPDIR=$PWD/.tmp .venv/bin/pytest tests/test_param_validation.py -q`

Result:
- `26 passed`

### Regression
Command:
- `TMPDIR=$PWD/.tmp .venv/bin/pytest tests/test_core.py tests/test_param_validation.py -q`

Result:
- `41 passed`

### Coverage (module)
Commands:
- `TMPDIR=$PWD/.tmp .venv/bin/coverage run -m pytest tests/test_param_validation.py -q`
- `.venv/bin/coverage report -m --include='*/tools/instantiate.py'`

Result:
- `src/sktime_mcp/tools/instantiate.py`: **98%**

## Note
In this runtime, `TMPDIR=$PWD/.tmp` was required for pytest due sandbox temp-dir restrictions.

### DIFF
--- a/src/sktime_mcp/tools/instantiate.py
+++ b/src/sktime_mcp/tools/instantiate.py
@@ -79,7 +79,7 @@ def _validate_params(
                 "warnings": warnings,
             }
 
-    # check if keys match known hyperparameters (warn, don't error)
+    # check if keys match known hyperparameters
     if estimator_name and params:
         registry = get_registry()
         node = registry.get_estimator_by_name(estimator_name)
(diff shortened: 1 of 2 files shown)


Shashankss1205:
instantiate_estimator and its params dict were replaced by instantiate(spec=...), which calls sktime's craft(); a typo'd key now fails immediately with the constructor's 'unexpected keyword argument' error, so the silent-failure path from #219 no longer exists. Closing as obsolete. A follow-up that appends the valid parameter list to that error would still be welcome as a small PR.

PR #258, opened 2026-04-20
fix: use <= in cleanup_old_jobs to handle zero-age boundary

hoo01:
## Problem
cleanup_old_jobs(max_age_hours=0) fails to remove a job created immediately before the call. The cutoff is computed as datetime.now(), and jobs created at that exact timestamp have created_at == cutoff, so the strict < comparison excludes them.

## Fix
Change < to <= so jobs with created_at == cutoff are also removed.

## Test
tests/test_background_jobs.py::test_cleanup_old_jobs now passes.

Applying to ESoC 2026 for the sktime agentic project. Happy to iterate based on maintainer feedback!

### DIFF
--- a/src/sktime_mcp/runtime/jobs.py
+++ b/src/sktime_mcp/runtime/jobs.py
@@ -305,7 +305,7 @@ class JobManager:
         cutoff = datetime.now() - timedelta(hours=max_age_hours)
 
         with self.lock:
-            old_job_ids = [job_id for job_id, job in self.jobs.items() if job.created_at < cutoff]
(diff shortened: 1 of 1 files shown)


Shashankss1205:
Verified: created_at == cutoff should be evicted when max_age_hours=0; the change still applies to current main and the suite passes. Not flaky on Linux CI but is on coarse-clock platforms. Merging.

Shashankss1205:
Landed on main as 23cfc7d (squash-merged locally because the org policy rejects the merge API for this token). Thanks!

PR #261, opened 2026-04-21
fix: correct cv_folds count and update deprecated import in evaluate …

hoo01:
## Problem
`evaluate_estimator` ignores the `cv_folds` parameter — the actual number of folds run is approximately `cv_folds * 2` due to a wrong formula in `initial_window` calculation.

## Fix
Changed `initial_window = max(int(n * 0.5), n - cv_folds * 2)` to `initial_window = max(1, n - cv_folds)`, which produces exactly `cv_folds` splits.

updated the deprecated `ExpandingWindowSplitter` import from `sktime.forecasting.model_selection` to `sktime.split`.

## Test
`tests/test_evaluate.py::test_evaluate_estimator_tool` now passes.

Applying to ESoC 2026 for the sktime agentic project. Happy to iterate based on maintainer feedback!

### DIFF
--- a/src/sktime_mcp/tools/evaluate.py
+++ b/src/sktime_mcp/tools/evaluate.py
@@ -8,7 +8,7 @@ import logging
 from typing import Any
 
 from sktime.forecasting.model_evaluation import evaluate
-from sktime.forecasting.model_selection import ExpandingWindowSplitter
+from sktime.split import ExpandingWindowSplitter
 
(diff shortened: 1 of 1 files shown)


Shashankss1205:
Duplicated Issue and PR as https://github.com/sktime/sktime-mcp/pull/208 and https://github.com/sktime/sktime-mcp/issues/207

Closing this

PR #262, opened 2026-04-21
fix: export_code generates private module import paths instead of pub…

hoo01:
## Bug
`export_code` uses `class.__module__` to resolve the import path, which returns
the private implementation module (e.g. `sktime.forecasting.naive._naive`)
instead of the public API path (`sktime.forecasting.naive`).

The exported code runs today, but private modules are not part of sktime's
public API contract — any internal refactor can silently break it.

## Fix
Added `_get_public_module()` helper that strips trailing underscore-prefixed
submodule components from the raw `__module__` string. Applied to both single
estimator and pipeline codegen paths.

## How I found it
Discovered while testing the full workflow (instantiate → fit_predict →
evaluate → export_code) through Claude Desktop with the sktime-mcp integration

as part of my ESOC 2026 application for the sktime Agentic Track.

### DIFF
--- a/src/sktime_mcp/tools/codegen.py
+++ b/src/sktime_mcp/tools/codegen.py
@@ -36,12 +36,20 @@ def _format_value(value: Any) -> str:
         return repr(value)
 
 
+def _get_public_module(module: str) -> str:
+    """Strip private submodule suffixes to get the public API import path."""
+    parts = module.split(".")
+    while parts and parts[-1].startswith("_"):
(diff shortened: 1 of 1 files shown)


Shashankss1205:
The bug was real. Since #488, export_code no longer emits per-estimator imports at all; it reproduces the handle from its craft() spec (codegen.py:190), so private module paths cannot appear, and the functions this PR patches were removed with that change. Closing as already resolved.

PR #263, opened 2026-04-21
feat: register MCP Resources and Prompts to expand protocol capability surface

### LINKED ISSUE #257: [ENH] Register MCP Resources and Prompts: sktime-mcp only uses one third of the MCP capability surface
# [ENH] Register MCP Resources and Prompts: sktime-mcp only uses one third of the MCP capability surface

## Problem

sktime-mcp exposes all its functionality through MCP Tools (22+ tools registered in `server.py`).
The MCP specification defines two other capability types, **Resources** and **Prompts**, neither of
which is registered anywhere in the codebase.

In practice this means:

- Every piece of sktime domain knowledge (estimator docs, tag definitions, dataset metadata) can
  only be retrieved by running a tool call. There is no way to read that context passively or cache
  it across turns.
- There are no reusable workflow templates. Each time an LLM needs to walk through a forecasting
  sequence, it reconstructs the steps from scratch.
- The SDK already supports both. `FastMCP` ships `@mcp.resource` and `@mcp.prompt` decorators
  (confirmed in the installed package), but nothing in the codebase uses them.

## Why This Matters

Resources and Prompts solve different problems than Tools, and each covers a real gap here.

**Resources** are for persistent, readable context that a client can pull on demand instead of
requesting through tool execution. Right now, an LLM trying to find estimators that support
prediction intervals has to call `list_estimators` with a tag filter, then `describe_estimator` for
each result. That burns multiple roundtrips on information that barely changes between sessions. A
resource at `sktime://estimators/{name}` would serve the same content without touching the tool
budget at all.

**Prompts** are reusable context injections. Without them, every session that starts with "forecast
this dataset" requires the model to rediscover the correct tool sequence: load data, instantiate an
estimator, call fit_predict, poll for completion, retrieve results. A prompt template makes the
intended workflow explicit and reduces the chance of the model doing something like calling
`fit_predict` before a data handle exists.

Clients that support MCP capability negotiation (Claude Desktop, Continue, Cursor) will actively use
Resources and Prompts when they are available. A server that only advertises tools is a lesser
integration for those clients. Resources and Prompts are in the spec for a reason, and skipping them
quietly limits what this server can do.

## Proposed Improvement

Register a focused set of resources and prompts that cover the most common information needs, without
duplicating what tools already do well.

**Resources (static or lightly-dynamic context):**

| URI | Purpose |
|-----|---------|
| `sktime://tags` | Tag catalog: name, description, allowed values, applicable tasks |
| `sktime://estimators/{name}` | Full documentation for a named estimator |
| `sktime://datasets` | Available demo datasets with short descriptions |

These map directly to data already computed by existing tools (`get_available_tags`,
`describe_estimator`, `list_datasets`). The resources expose the same content through a resource-read
path rather than a tool-call path.

**Prompts (workflow templates):**

| Name | Purpose |
|------|---------|
| `forecast_univariate` | Loading data, picking a forecaster, running fit_predict |
| `compare_forecasters` | Side-by-side evaluation of two or more estimators |
| `build_pipeline` | Composing transformers and a final estimator |

These do not execute anything. They inject structured guidance so the model knows which tools to call
and in what order.

## Example Use Case

An LLM is asked to forecast sales data. With the current server:
1. Call `list_estimators` to see what's available.
2. Call `describe_estimator` one or more times to understand capabilities.
3. Infer the correct tool sequence from scratch.

With a `forecast_univariate` prompt, the client can inject a pre-written workflow outline into
context upfront. With `sktime://estimators/{name}` as a resource, it reads estimator documentation
without a tool call. Fewer roundtrips, shorter conversation, less room for ordering mistakes.

## Possible Direction

The `FastMCP` decorator API makes this low-effort to add. A resource registration:

```python
@mcp.resource("sktime://estimators/{name}")
def estimator_resource(name: str) -> str:
    result = describe_estimator_tool(name)
    if not result["success"]:
        raise ValueError(f"Unknown estimator: {name}")
    return json.dumps(result, indent=2)
```

A prompt registration:

```python
@mcp.prompt(name="forecast_univariate")
def forecast_prompt() -> list[dict]:
    return [
        {
            "role": "user",
            "content": "Walk me through forecasting a univariate time series with sktime-mcp."
        },
        {
            "role": "assistant",
            "content": (
                "To forecast a univariate series:\n"
                "1. Load your data with `load_data_source`.\n"
                "2. Pick an estimator with `list_estimators` (filter by task: forecasting).\n"
                "3. Instantiate it with `instantiate_estimator`.\n"
                "4. Run `fit_predict` with your data handle and horizon.\n"
                "5. Poll `check_job_status` until complete, then call `get_job_result`."
            )
        }
    ]
```

No new modules needed for an initial implementation. Both registrations can live in `server.py`
alongside the existing tool registrations, using the same `executor` and `describe_estimator_tool`
functions that tools already call.

## Alternatives Considered

**Expanding tool descriptions instead:** Tool descriptions in the JSON schema are the existing
workaround, but they only appear in the tool list, get truncated by most clients, and cannot carry
multi-turn workflow guidance. They are also not queryable by URI or cacheable.

**Leaving it as-is:** Fine for clients that only use tools, but misses a real opportunity for clients
(Claude Desktop, Continue, Cursor) that will surface resources and prompts directly in their UI when
the server advertises them.

## Additional Context

- FastMCP resource signature: `@mcp.resource(uri, *, name, description, mime_type)`
- FastMCP prompt signature: `@mcp.prompt(name, description)`
- Both decorators confirmed present in the installed SDK (`mcp.server.fastmcp.FastMCP`)
- Closed issue #30 asked for richer tag metadata but was considered too verbose for tool output.
  The `sktime://tags` resource is a natural home for that content.
- MCP specification covers Resources and Prompts as first-class capability types alongside Tools

I'd be happy to start working on this and submit a PR for the same.
@Shashankss1205

hoo01:
## Problem
sktime-mcp exposes all functionality through MCP Tools only. The MCP specification defines two other capability types — Resources and Prompts — neither of which was registered in the server.

Closes #257

## Changes
**Resources** (passive readable context, no tool budget consumed):
- `sktime://tags` — full tag catalog with descriptions and allowed values
- `sktime://datasets` — available demo datasets
- `sktime://estimators/{name}` — full documentation for any named estimator (URI template)

**Prompts** (reusable workflow templates injected into client context):
- `forecast_univariate` — step-by-step univariate forecasting workflow
- `compare_forecasters` — side-by-side evaluating multiple forecasters
- `build_pipeline` — composing transformers and a forecaster into a pipeline

## Implementation
Uses the low-level `mcp.server.Server` decorator API (`@server.list_resources`, 
`@server.list_resource_templates`, `@server.read_resource`, `@server.list_prompts`, 
`@server.get_prompt`). All resource content is served by reusing existing tool 
functions — no logic duplication.

Verified: all 8 handlers registered correctly on server startup.

## Notes
Claude Desktop currently surfaces only Tools in its UI — Resources and Prompts are available at the protocol level and will be used by clients that support MCP capability negotiation (Cursor, Continue, etc.).

Implemented as part of my ESOC 2026 application for the sktime Agentic Track.

### DIFF
--- a/src/sktime_mcp/server.py
+++ b/src/sktime_mcp/server.py
@@ -807,7 +807,7 @@ async def read_resource(uri: AnyUrl) -> list[TextResourceContents]:
 
     if uri_str == "sktime://tags":
         result = get_available_tags()
-        text = json.dumps(result, indent=2)
+        text = json.dumps(result.get("tags", result), indent=2)
     elif uri_str == "sktime://datasets":
         from sktime_mcp.runtime.executor import get_executor
         executor = get_executor()
@@ -816,7 +816,7 @@ async def read_resource(uri: AnyUrl) -> list[TextResourceContents]:
     elif uri_str.startswith("sktime://estimators/"):
         name = uri_str[len("sktime://estimators/"):]
         result = describe_estimator_tool(name)
-        text = json.dumps(result, indent=2)
+        text = json.dumps({k: v for k, v in result.items() if k != "success"}, indent=2)
     else:
         raise ValueError(f"Unknown resource URI: {uri_str}")
 
--- a/src/sktime_mcp/server.py
+++ b/src/sktime_mcp/server.py
@@ -13,7 +13,17 @@ from typing import Any
 
 from mcp.server import Server
 from mcp.server.stdio import stdio_server
-from mcp.types import TextContent, Tool
+from mcp.types import (
+    GetPromptResult,
+    Prompt,
+    PromptMessage,
+    Resource,
+    ResourceTemplate,
+    TextContent,
+    TextResourceContents,
+    Tool,
+)
+from pydantic import AnyUrl
 
 from sktime_mcp.composition.validator import get_composition_validator
 from sktime_mcp.tools.codegen import export_code_tool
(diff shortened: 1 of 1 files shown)


Om-Dhondge:
Hey @hoo01, one thing worth flagging: the existing server uses FastMCP throughout. This PR drops down to the low-level ```mcp.server.Server``` API for the new handlers. Is there a reason to bypass the ```FastMCP``` decorator approach here? The ```@mcp.resource``` and ```@mcp.prompt``` decorators are available in the installed SDK and would keep the registration style consistent with the rest of ```server.py```.

hoo01:
> Hey @hoo01, one thing worth flagging: the existing server uses FastMCP throughout. This PR drops down to the low-level `mcp.server.Server` API for the new handlers. Is there a reason to bypass the `FastMCP` decorator approach here? The `@mcp.resource` and `@mcp.prompt` decorators are available in the installed SDK and would keep the registration style consistent with the rest of `server.py`.

Thanks for the review! Just to clarify, the server actually uses the low-level mcp.server.Server API (in server.py, from mcp.server import Server, server = Server("sktime-mcp")), not FastMCP, so the new handlers follow the same pattern as the existing @server.list_tools() and @server.call_tool() registrations.

Happy to discuss further if anything looks off!

Om-Dhondge:
Thanks for clarifying, you're right. I missed that the server uses the ```Server``` API directly. The implementation makes sense then. Apologies for the confusion.

Om-Dhondge (commented):
One small thing: ```sktime://tags``` serializes the raw return value of ```get_available_tags()```, which wraps the content in ```{"success": true, "tags": {...}}```. A client reading the resource receives that envelope instead of the tag data directly. Same issue applies to the estimator resource. The fix would be to extract just ```result["tags"]``` / the estimator fields before serializing, since resources are read as content rather than called as tools.

hoo01:
> One small thing: `sktime://tags` serializes the raw return value of `get_available_tags()`, which wraps the content in `{"success": true, "tags": {...}}`. A client reading the resource receives that envelope instead of the tag data directly. Same issue applies to the estimator resource. The fix would be to extract just `result["tags"]` / the estimator fields before serializing, since resources are read as content rather than called as tools.

thanks for your catching that!!! have fixed it , now return the content directly.

Shashankss1205:
The approach (low-level @server.list_resources/@server.get_prompt handlers on the existing Server instance) fits server.py and #257 is still open. However, the content is entirely tied to the pre-consolidation API: read_resource references get_available_tags, describe_estimator_tool and executor.list_datasets() (all removed; ruff flags two undefined names), and every prompt tells the client to call list_estimators/instantiate_estimator/fit_predict/evaluate_estimator/validate_pipeline. Since essentially every line would change, closing; a fresh PR against the current 26-tool surface with tests for read_resource/get_prompt would be very welcome.

PR #266, opened 2026-04-21
feat: add fit_transform and transform MCP tools for sktime transformers

### LINKED ISSUE #167: Add standalone fit_transform / transform MCP tools for sktime transformers
## Is your feature request related to a problem? Please describe.

Currently the MCP server can instantiate transformer pipelines via `instantiate_pipeline`, but there is no dedicated MCP tool to execute a single `sktime.transformations` transformer on data. This means LLM-driven workflows are currently limited to full model pipelines and cannot use sktime purely for preprocessing or feature extraction.

## Describe the solution you'd like

Add dedicated MCP tools for transformer execution:

- `fit_transform`
- `transform`

These tools should:
- accept a transformer handle created by `instantiate_estimator` or `instantiate_pipeline`
- accept either demo dataset names or loaded data handles
- call the transformer's `fit_transform` or `transform` methods
- return JSON-safe transformed data
- preserve the existing handle-based runtime model where Python objects stay server-side

Potential implementation points:
- add new tool definitions in `src/sktime_mcp/server.py`
- add tool implementations in `src/sktime_mcp/tools/`
- add executor support in `src/sktime_mcp/runtime/executor.py`
- support transformer-only workflows without requiring a downstream predictor

## Additional context

This feature would enable a new class of LLM workflows:
- data preprocessing and cleaning with `sktime` transformers
- feature extraction from time series
- applying detrending, deseasonalizing, smoothing, and other transformations before later modeling
- using `sktime` as a transformation engine, not only a forecasting/classification engine

tarun-227:
## Summary
- Add `fit_transform` tool: fits a sktime transformer on data and returns the transformed result in one step; marks the handle as fitted
- Add `transform` tool: applies an already-fitted transformer to new data without re-fitting; returns a clear error if called before `fit_transform`
- Both tools accept either a demo dataset name or a loaded data handle
- Register both in `list_tools()` and dispatch in `call_tool()`
- 7 new tests covering success paths, fitted-state tracking, error cases, and server tool exposure

## Context
Closes #167. sktime-mcp had no way to run standalone transformer workflows. LLMs could only use transformers as part of a forecasting pipeline via `instantiate_pipeline`. With `fit_transform` and `transform` exposed as MCP tools, an LLM can now:
- detrend or deseasonalize a series before modelling
- apply log or Box-Cox scaling as a preprocessing step
- extract features from raw time series
- apply the same fitted transformation to held-out data via `transform`

## Test plan
- [x] `test_fit_transform_demo_dataset` — returns transformed series with correct shape
- [x] `test_fit_transform_marks_handle_fitted` — handle is marked fitted after call
- [x] `test_transform_after_fit_transform` — transform succeeds on a fitted handle
- [x] `test_transform_without_fit_fails` — clear error if transform called before fit_transform
- [x] `test_fit_transform_no_data_source_errors` — error when no data source given
- [x] `test_fit_transform_unknown_handle_errors` — error on missing handle
- [x] `test_server_exposes_transform_tools` — both tools appear in list_tools()
- [x] All 22 tests pass (15 pre-existing + 7 new), 0 regressions
- [x] `ruff check` clean

AI was used for basic drafting assistance and refinement; all changes were verified manually.


Shashankss1205:
Since the refactor, transformers are first-class in the generic tools: instantiate('Detrender()') -> fit -> predict (which calls transform for transformer handles), plus transform_data for one-shot transforms and call_method for anything else, so separate fit_transform/transform tools would duplicate that (#270). The PR also edits instantiate.py, which was rewritten around craft(). Closing as superseded; #167 is closed too.

PR #269, opened 2026-04-22
fix: validate max_age_hours >= 1 in cleanup_old_jobs_tool to prevent silent mass-deletion

### LINKED ISSUE #249: [BUG] cleanup_old_jobs_tool accepts zero/negative max_age_hours silently deletes all jobs
**Describe the bug**

**To Reproduce**

```

# Create some jobs first
job_manager.create_job(
    job_type="fit_predict",
    dataset_name="airline",
    total_steps=3,
)
job_manager.create_job(
    job_type="fit_predict",
    dataset_name="sunspots",
    total_steps=3,
)

print(result)

```

**Expected behavior**

**Additional context**

```
Proposed fix (2 lines added):
    if max_age_hours < 1:
```

amethystani:
Fixes #249.

`cleanup_old_jobs_tool` passed `max_age_hours` directly to `job_manager.cleanup_old_jobs()` without any validation. Because the cutoff is computed as `datetime.now() - timedelta(hours=max_age_hours)`:

- `max_age_hours=0` yields a cutoff equal to the current time, so every job looks "old" and is deleted silently.
- Any negative value pushes the cutoff into the future, again matching every job and wiping the entire store.

Neither case produced an error or any indication that the operation was unsafe.

**Before** (broken):

```python
# → {"success": True, "message": "Removed 2 old job(s)", "count": 2}
# All jobs gone, no warning.
```

**After** (fixed):

```python
# → {"success": False, "error": "max_age_hours must be a positive integer (>= 1)."}
```

The guard is added only in the MCP-facing tool. The underlying `job_manager.cleanup_old_jobs(0)` behaviour is unchanged — existing internal tests rely on it for teardown.

Three regression tests are added covering:
- `max_age_hours=0` → rejected
- `max_age_hours=-5` → rejected
- `max_age_hours=1` → accepted

### DIFF
--- a/src/sktime_mcp/tools/job_tools.py
+++ b/src/sktime_mcp/tools/job_tools.py
@@ -118,11 +118,18 @@ def cleanup_old_jobs_tool(max_age_hours: int = 24) -> dict[str, Any]:
     Clean up old jobs.
 
     Args:
-        max_age_hours: Maximum age in hours (default: 24)
+        max_age_hours: Maximum age in hours for a job to be considered old (default: 24).
+            Must be a positive integer (>= 1).
 
     Returns:
         Dictionary with number of jobs removed
     """
+    if max_age_hours < 1:
+        return {
+            "success": False,
+            "error": "max_age_hours must be a positive integer (>= 1).",
+        }
+
     job_manager = get_job_manager()
 
     count = job_manager.cleanup_old_jobs(max_age_hours)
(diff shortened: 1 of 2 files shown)


amethystani:
@fkiraly — tagging you as the sktime-mcp maintainer. The fix adds a `max_age_hours < 1` guard in `cleanup_old_jobs_tool` so zero/negative values return an error instead of silently wiping every job in the store. Happy to adjust the threshold or error format if preferred.

Shashankss1205:
cleanup_old_jobs_tool was removed in the tool consolidation; job cleanup is now an internal periodic task driven by SKTIME_MCP_JOB_MAX_AGE_HOURS, with no tool-level max_age_hours argument left to validate. Closing as obsolete (same as #274). Validating the env var in config.py is tracked in #249.

Issue #221, opened 2026-04-17
[BUG] list_estimators_tool accepts negative limit — returns 400+ results instead of validation error


harsh2025-sketch:
## Summary

`list_estimators_tool` validates `offset >= 0` but does NOT validate `limit`.
Passing a negative `limit` (e.g. `-5`) causes Python slice semantics to return 
almost all estimators instead of raising an error. This can flood an LLM agent 
with hundreds of results, consuming its entire context window.

## Steps to Reproduce

```python
import sys
sys.path.insert(0, "src")
from sktime_mcp.tools.list_estimators import list_estimators_tool

result = list_estimators_tool(limit=-5, offset=0)

print("success:", result["success"])   # True  (should be False)
print("count:", result["count"])       # 410   (should be error)
print("total:", result["total"])       # 415
```

## What's Happening

```python
# list_estimators.py — only offset is validated, not limit
if offset < 0:
    return {"success": False, "error": "offset must be non-negative"}
# limit is never checked!

estimators = all_estimators[offset : offset + limit]
# limit=-5 → all_estimators[0:-5] → returns 410 items!
```

## Expected vs Actual

| Input | Expected | Actual |
|---|---|---|
| `limit=-5` | Validation error | 410 results returned |
| `limit=-1` | Validation error | All but 1 result returned |

## Suggested Fix

```python
# Add this check before the slice:
if limit is not None and limit < 0:
    return {
        "success": False,
        "error": "limit must be a non-negative integer"
    }
```

Add tests:
```python
def test_negative_limit_returns_error():
    result = list_estimators_tool(limit=-1, offset=0)
    assert result["success"] is False

def test_zero_limit_returns_empty():
    result = list_estimators_tool(limit=0, offset=0)
    assert result["count"] == 0
```

## Environment
- sktime-mcp: 0.1.0
- Python: 3.12.9
- sktime: 0.40.1
- Reproduced via single-process run

harsh2025-sketch:
Found this during codebase audit for ESoC 2026 (sktime agentic track).
Fix is a 3-line guard + tests. Happy to submit a PR immediately.

PR #274, opened 2026-04-22
[BUG] Fix cleanup_old_jobs_tool accepting zero/negative max_age_hours (#249)

### LINKED ISSUE #249: [BUG] cleanup_old_jobs_tool accepts zero/negative max_age_hours silently deletes all jobs

krishnachaudhary18:
#### Reference Issues/PRs
Fixes #249

The cleanup_old_jobs_tool was silently deleting all jobs when someone passed max_age_hours=0 or a negative value like -1. This happened because datetime.now() - timedelta(hours=0) gives the current time, so every job looked "old" and got wiped out without any warning.

I added a simple validation check at the top of cleanup_old_jobs_tool() if max_age_hours is less than 1, the tool now returns a clear error message instead of proceeding with the deletion. I also updated the docstring to document this behavior.

No, no new dependencies added.

- The validation guard in src/sktime_mcp/tools/job_tools.py (lines 129–133) just 2 lines of logic plus the error return.
- The updated test_cleanup_old_jobs test in tests/test_background_jobs.py the old version actually relied on the bug (max_age_hours=0) to work, so I fixed it to backdate a job properly instead.
- The 3 new test functions that cover zero, negative, and valid values.

#### Any other comments?
##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).

### DIFF
--- a/src/sktime_mcp/tools/job_tools.py
+++ b/src/sktime_mcp/tools/job_tools.py
@@ -118,11 +118,20 @@ def cleanup_old_jobs_tool(max_age_hours: int = 24) -> dict[str, Any]:
     Clean up old jobs.
 
     Args:
-        max_age_hours: Maximum age in hours (default: 24)
+        max_age_hours: Maximum age in hours. Must be a positive integer (>= 1).
+            Passing 0 or a negative value would set the cutoff to "now" or the
+            future, causing all jobs to be silently deleted. (default: 24)
 
     Returns:
-        Dictionary with number of jobs removed
+        Dictionary with number of jobs removed, or an error dict if the
+        argument is invalid.
     """
+    if max_age_hours < 1:
+        return {
+            "success": False,
+            "error": "max_age_hours must be a positive integer (>= 1).",
+        }
+
     job_manager = get_job_manager()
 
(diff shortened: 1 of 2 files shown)


krishnachaudhary18:
@Shashankss1205,

Shashankss1205:
The tool it guards, cleanup_old_jobs_tool, was removed in the refactor; cleanup now runs as a periodic server task using SKTIME_MCP_JOB_MAX_AGE_HOURS (server.py _periodic_job_cleanup), so an LLM can no longer pass 0 or a negative age. Closing as obsolete. The remaining gap (the env var itself is not validated, config.py) is tracked in #249.

PR #275, opened 2026-04-23
fix: handle NumPy/Pandas types in sanitize_for_json (#197)

### LINKED ISSUE #197: [BUG] sanitize_for_json crashes when returning NumPy or Pandas types

SarthakB11:
## Summary

Fixes #197: `sanitize_for_json` crashes when tool results contain NumPy or Pandas types, causing `json.dumps` to raise `TypeError` on the MCP response.

**Root cause:** The function had no handling for numpy/pandas types — they fell through to `else: return obj`, which `json.dumps` cannot serialize.

**Note:** Existing PRs #239 and #240 only added circular-reference detection (fixing #189) but never addressed the numpy/pandas type serialization that is the actual subject of #197.

## Changes

**`src/sktime_mcp/server.py`** — rewrites `sanitize_for_json`:
- `np.bool_` → `bool`, `np.integer` → `int`, `np.floating` → `float` (NaN/inf → `None`)
- `np.ndarray` → list via `.tolist()`
- `pd.Timestamp` → ISO 8601 string
- `pd.NaT`, `pd.NA` → `None`
- `pd.Series` → list, `pd.DataFrame` → list of dicts
- Cycle detection retained but scoped to **containers only** — the previous approach tracked scalar `id()`s which can give false positives due to CPython's small-int cache (e.g. `{"a": 42, "b": 42}` would incorrectly replace the second `42` with `"<circular reference>"`)
- `_seen.discard()` after processing each container so the same object can appear in sibling branches without false detection

**`tests/test_sanitize_for_json.py`** — 25 new tests covering:
- Native Python types (no regression)
- All NumPy scalar/array types including NaN and inf edge cases
- All Pandas types (Timestamp, NaT, NA, Series, DataFrame)
- Circular reference detection
- False-positive checks for sibling objects and small-int caching

## Test results

```
25 passed in 0.61s
```

## Scope

Only `src/sktime_mcp/server.py` (the `sanitize_for_json` function) and the new test file — no other files changed.

### DIFF
--- a/src/sktime_mcp/server.py
+++ b/src/sktime_mcp/server.py
@@ -67,16 +67,79 @@ logger = logging.getLogger(__name__)
 server = Server("sktime-mcp")
 
 
-def sanitize_for_json(obj):
-    """Recursively convert objects to JSON-serializable format."""
-    if isinstance(obj, dict):
-        return {str(k): sanitize_for_json(v) for k, v in obj.items()}
-    elif isinstance(obj, (list, tuple)):
-        return [sanitize_for_json(item) for item in obj]
-    elif hasattr(obj, "__dict__") and not isinstance(obj, (str, int, float, bool, type(None))):
-        return str(obj)
-    else:
+def sanitize_for_json(obj: object, _seen: set[int] | None = None) -> object:
+    """Recursively convert objects to JSON-serializable format.
+
+    Handles NumPy scalars/arrays and Pandas types in addition to plain Python
+    containers. Cycle detection is applied to containers only to avoid false
+    positives with CPython's small-int caching.
+    """
+    # --- NumPy types ---
+    try:
+        import numpy as np
+
+        if isinstance(obj, np.bool_):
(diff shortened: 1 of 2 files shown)


Shashankss1205:
Already merged in a previous PR

Issue #270, opened 2026-04-22
[ENH] consolidate list of tools


fkiraly:
Currently, there is a large set of tools (24), and about that many suggestions for adding new tools as well.

We need to come up with a consolidated - and non-duplicative - set of tools that we think is covering common use cases well.

For that, we should have a list of tools (with signature, options, etc), as well as use cases they address, next to each other.

This can be based on the set of current tools and current ideas in pull requests.

This issue is solved when we have:

* a list of tools with signatures and meaning (agreed by maintainers discussing here)
* a list of use cases or references to them, from the tools, to clarify in which use cases they would be called

The tools themselves need not be implemented or merged yet, first we need to sharpen the design.

ISSUE FOR MAINTAINERS ONLY - NOT AN ISSUE FOR NEW CONTRIBUTORS

Shashankss1205:
# sktime-mcp Tool Reference

## Discovery

| Tool | Args | What it does |
|------|------|--------------|
| `list_estimators` | `task`, `tags`, `query`, `limit`, `offset` | Search and filter estimators from the sktime registry |
| `describe_estimator` | `estimator` | Full docs, params, tags, and import path for a named estimator |
| `get_available_tags` | — | All queryable tags with descriptions and valid values |

## Instantiation

| Tool | Args | What it does |
|------|------|--------------|
| `instantiate_estimator` | `estimator`, `params` | Create an estimator instance with given hyperparams |
| `instantiate_pipeline` | `components`, `params_list` | Compose and validate a pipeline from named steps in one shot |
| `list_handles` | — | List all active estimator handles in memory |
| `release_handle` | `handle` | Free an estimator handle from memory |

## Execution

| Tool | Args | What it does |
|------|------|--------------|
| `fit` | `estimator_handle`, `estimator_type`, `dataset`, `data_handle` | Fit any estimator (forecaster, classifier, clusterer, detector, annotator) and store fitted state under the handle |
| `predict` | `estimator_handle`, `mode`, `fh`, `coverage`, `alpha` | Run prediction on a fitted handle — `mode` selects `predict` / `predict_interval` / `predict_quantiles` / `predict_proba` / `predict_scores` |
| `transform` | `estimator_handle`, `data_handle`, `action` | Apply a transformer — `action` selects `fit_transform` / `transform` / `inverse_transform` |
| `update` | `estimator_handle`, `y_new` | Update a fitted model with new incoming data for streaming or online use |
[...]

PR #276, opened 2026-04-23
Add profile timeseries tool

sArThAk1612:
Motivation
LLM-based workflows need a compact, interpretable summary of a time series
to make decisions (e.g. model selection) without directly accessing raw data.

Currently, sktime-mcp does not provide a standardized way to extract such
characteristics.

What this PR does

This PR adds a new MCP tool `profile_time_series` that computes a lightweight
set of descriptive statistics for a time series, including:

 length and missing ratio
mean and standard deviation
 trend proxy (correlation with time)
 seasonality proxy (autocorrelation)
 stationarity heuristic
 recommended forecasting horizon

Why this approach
I chose simple heuristics instead of heavy statistical tests to keep the tool:

 fast
 interpretable
 easy for LLMs to consume
How I tested it

 Added unit test for output structure
 Verified manually on simple increasing series

Future direction

This tool can serve as a first step in a pipeline like:

profile_time_series - model recommendation -evaluation

### DIFF
--- a/examples/csv_example.py
+++ b/examples/csv_example.py
@@ -4,7 +4,7 @@ Example: Loading data from a CSV file.
 This example demonstrates how to use the sktime-mcp data loading
 functionality with CSV files.
 """
-
+import tempfile
 from pathlib import Path
 
 import pandas as pd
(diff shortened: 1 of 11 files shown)


Shashankss1205:
This PR does not register the tool in server.py, bundles the router/trace/diagnostics files from your other branches, and adds test and example scripts inside the package. inspect_data already reports shape, frequency, missing values and summary stats, and maintainers are consolidating tools rather than adding heuristic profilers (#270). Closing.

PR #285, opened 2026-04-23
[BUG] Reject unfitted estimators in `save_model_tool` before MLflow call

### LINKED ISSUE #280: [BUG] `save_model_tool` silently attempts to save unfitted estimators

`save_model_tool` resolves the estimator from a handle and immediately calls `sktime.utils.mlflow_sktime.save_model` without checking whether the estimator has been fitted:

\`\`\`python
def save_model_tool(estimator_handle, path, mlflow_params=None):
    handle_manager = get_handle_manager()
    estimator = handle_manager.get_instance(estimator_handle)  # no fitted check
    ...
    save_model(sktime_model=estimator, path=path, ...)
\`\`\`

#### Impact

If an LLM calls `instantiate_estimator → save_model` without an intermediate `fit_predict`, it will save an unfitted sktime object. An unfitted sktime estimator cannot `predict`; loading the saved model and calling `predict` will raise a `NotFittedError`. This is particularly confusing because:

1. `save_model_tool` returns `{"success": true, ...}` — no indication anything is wrong.
2. The error only surfaces when the model is *loaded* and used later, potentially in a different context.
3. The `HandleManager.is_fitted()` method already exists and is used in `predict` — it is simply not wired into `save_model_tool`.

#### Steps to reproduce

\`\`\`python
from sktime_mcp.tools.save_model import save_model_tool

result = instantiate_estimator_tool("NaiveForecaster")
handle = result["handle"]

# Save without fitting — currently succeeds silently
save_result = save_model_tool(handle, "/tmp/model_dir")
assert save_result["success"]  # True — but model is unfitted
\`\`\`

#### Expected behavior

`save_model_tool` should check `handle_manager.is_fitted(estimator_handle)` and return a descriptive error when the estimator has not been fitted, rather than serializing an unusable object.

\`\`\`python
{"success": False, "error": "Estimator 'NaiveForecaster' (handle: est_abc) has not been fitted. Run fit_predict before saving."}
\`\`\`

#### Proposed fix

Add a pre-flight `is_fitted` check in `save_model_tool` before the MLflow call. One line check + one early return.

#### Affected files

- `src/sktime_mcp/tools/save_model.py`

#### PR checklist
- [ ] I've added unit tests and made sure they pass locally.

magic-peach:
#### Reference Issues/PRs

Closes #280

`save_model_tool` previously serialised estimator handles unconditionally — including ones that had never been fitted. The resulting artefact raises `NotFittedError` when loaded and asked to predict, and the only signal came from MLflow deep in the call stack, making it hard to diagnose.

This PR adds a two-line pre-flight check immediately after the handle is resolved:

```python
if not handle_manager.is_fitted(estimator_handle):
    handle_info = handle_manager.get_info(estimator_handle)
    return {
        "success": False,
        "error": (
            f"Estimator '{handle_info.estimator_name}' (handle: {estimator_handle}) "
            "has not been fitted. Call fit_predict before saving."
        ),
    }
```

`HandleManager.is_fitted()` already existed and was used by the `predict` path — this change simply wires it into `save_model_tool`.

**Changed files:**

| File | Change |
|------|--------|
| `src/sktime_mcp/tools/save_model.py` | Add `is_fitted` guard before the MLflow call |
| `tests/test_save_model_fitted_check.py` | Three new tests: unfitted → error, unknown handle → error, fitted → proceeds to save |

No.

- Wording of the error message — should it reference `fit_predict` specifically, or the lower-level `fit` call as well?
- Whether the check should be a warning rather than a hard error (i.e. allow saving unfitted models with an explicit flag). Current opinion: hard error is safer and consistent with the predict path.

#### Any other comments?

The existing `test_save_model_tool` test in `test_core.py` already tests the happy path via `mark_fitted`; the new tests cover the new guard and the unchanged unknown-handle path.

##### For all contributions

### DIFF
--- a/src/sktime_mcp/tools/save_model.py
+++ b/src/sktime_mcp/tools/save_model.py
@@ -44,6 +44,16 @@ def save_model_tool(
     except KeyError:
         return {"success": False, "error": f"Handle not found: {estimator_handle}"}
 
+    if not handle_manager.is_fitted(estimator_handle):
+        handle_info = handle_manager.get_info(estimator_handle)
+        return {
+            "success": False,
+            "error": (
+                f"Estimator '{handle_info.estimator_name}' (handle: {estimator_handle}) "
+                "has not been fitted. Call fit_predict before saving."
+            ),
+        }
+
     if mlflow_params is not None and not isinstance(mlflow_params, dict):
         return {"success": False, "error": "mlflow_params must be a dictionary"}
 
--- a/tests/test_save_model_fitted_check.py
+++ b/tests/test_save_model_fitted_check.py
@@ -1,0 +1,61 @@
+"""Tests for the fitted-check guard added to save_model_tool."""
+
+import sys
+
+sys.path.insert(0, "src")
+
+
+class TestSaveModelFittedCheck:
+    """Tests that save_model_tool rejects unfitted estimators early."""
+
+    def test_save_unfitted_estimator_returns_error(self):
+        """Saving an unfitted handle must fail with a clear error message."""
+        from sktime_mcp.runtime.handles import get_handle_manager
+        from sktime_mcp.tools.save_model import save_model_tool
+
+        handle_manager = get_handle_manager()
+        handle = handle_manager.create_handle("NaiveForecaster", object())
+
+        try:
+            result = save_model_tool(handle, "/tmp/should_not_be_created")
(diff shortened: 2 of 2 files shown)


Shashankss1205 (approved):
Good one!

PR #286, opened 2026-04-23
[ENH] `export_code` uses training dataset from handle metadata

### LINKED ISSUE #281: [ENH] `export_code` with `include_fit_example=True` silently falls back to 'airline' dataset

When `export_code` is called with `include_fit_example=True` and no `dataset` argument, it silently falls back to the `"airline"` dataset regardless of what data the model was actually trained on:

\`\`\`python
# tools/codegen.py
if dataset and dataset in DEMO_DATASETS:
    ...
else:
    # Default to airline for backward compatibility
    loader_module = "sktime.datasets"
    loader_func = "load_airline"
\`\`\`

#### Impact

A typical LLM flow is:

1. `instantiate_estimator("ARIMA")`
2. `fit_predict(estimator_handle=..., dataset="sunspots")`
3. `export_code(handle=..., include_fit_example=True)`

Step 3 generates code that loads `airline` data — *not* `sunspots`. The exported example is immediately wrong and misleading. This erodes trust in generated code and forces the user to manually fix the dataset reference.

The information needed to fix this already flows through the system: `executor.fit_predict` knows which dataset was used. The `HandleInfo` dataclass has a `metadata: dict` field intended for exactly this kind of supplementary information. It just is not populated.

#### Proposed fix

1. **In `executor.fit_predict`**: after a successful fit, store the dataset name in the handle's metadata:
   \`\`\`python
   handle_info = self._handle_manager.get_info(handle_id)
   handle_info.metadata["training_dataset"] = dataset
   \`\`\`

2. **In `export_code_tool`**: when `include_fit_example=True` and no `dataset` argument is given, read the training dataset from handle metadata as the default, before falling back to `"airline"`:
   \`\`\`python
   effective_dataset = dataset or handle_info.metadata.get("training_dataset") or "airline"
   \`\`\`

This is a non-breaking change — `dataset` passed explicitly always takes priority, and the `"airline"` fallback is kept for handles that were created before this fix or never trained.

#### Affected files

- `src/sktime_mcp/runtime/executor.py` — write `training_dataset` to handle metadata on fit
- `src/sktime_mcp/tools/codegen.py` — read `training_dataset` from handle metadata as default

#### PR checklist

magic-peach:
#### Reference Issues/PRs

Closes #281

`export_code` with `include_fit_example=True` previously always fell back to the `"airline"` dataset unless the caller explicitly passed a `dataset` argument. This meant:

```python
fit_predict(handle, dataset="sunspots")   # trained on sunspots
export_code(handle, include_fit_example=True)  # generated code loads airline — wrong!
```

The fix is a two-part change:

1. **`executor.fit_predict`** — after a successful fit, write the dataset name into the handle's existing `metadata` dict under `"training_dataset"`.
2. **`export_code_tool`** — resolve the effective dataset via a three-level priority chain:
   - Explicit `dataset` argument (unchanged behaviour)
   - `handle_info.metadata.get("training_dataset")` ← **new**
   - `"airline"` (backward-compatible fallback for unfitted or pre-change handles)

**Changed files:**

| File | Change |
|------|--------|
| `src/sktime_mcp/runtime/executor.py` | Write `metadata["training_dataset"] = dataset` after successful fit |
| `src/sktime_mcp/tools/codegen.py` | Read `training_dataset` from handle metadata as the default before falling back to `"airline"` |
| `tests/test_export_code_dataset_tracking.py` | Four tests: metadata stored after fit_predict; export uses it; explicit arg overrides; unfitted handle falls back to airline |

No.

- The metadata write in `executor.fit_predict` uses `handle_info.metadata["training_dataset"] = dataset`. `HandleInfo.metadata` is a plain `dict` — no locking needed since the MCP server is single-threaded in practice. Confirm this is safe.
- The `data_handle` path in `fit_predict` does not record a training_dataset (there is no single string name for a custom loaded dataset). This is intentional; the fallback to `"airline"` is still reasonable for custom-data handles.

#### Any other comments?

`fit_predict_async` does not forward the dataset name to metadata yet; that would be a natural follow-up.

##### For all contributions


Shashankss1205 (approved):
Thanks!

PR #296, opened 2026-04-25
Add Windows + Python 3.14 scipy installation workaround to README

sArThAk1612:
PROBLEM
Installing with `pip install -e ".[all]"` fails on Windows with Python 3.14
because scipy attempts to build from source and requires C17, which is not
supported by MSVC.

This occurs when pre-built wheels for scipy are not available.

Solution

Added a note in the README installation section explaining the issue and
providing a workaround using:

    pip install scipy --only-binary=:all:
    pip install -e "."

This ensures users install pre-built binaries and avoid compilation errors.

Why this helps

Improves installation experience for Windows users on newer Python versions
and prevents confusing build failures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,18 @@ python3 -m pip install -e .
 
 # With all optional dependencies
 python3 -m pip install -e ".[all]"
+### ⚠️ Windows + Python 3.14 note
+
+On Windows with Python 3.14, installing with:
+    pip install -e ".[all]"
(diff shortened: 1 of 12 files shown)


Shashankss1205:
The PR seems to be polluted with a different code change, Can you please verify this once? @sArThAk1612 

Also everything from the examples are just literal examples. We dont ship it as part of the repo.

sArThAk1612:
yup ill verify and get back

PR #297, opened 2026-04-25
Add coverage parameter validation in fit_predict_tool

sArThAk1612:
Problem

The `fit_predict_tool` currently does not accept or validate a `coverage`
parameter for probabilistic forecasting workflows.

As a result, if coverage support is introduced or used by downstream tools,
invalid inputs (e.g., strings or values outside (0,1)) could lead to
unexpected behavior or runtime errors.

Solution

This PR introduces an optional `coverage` parameter to `fit_predict_tool`
and adds input validation to ensure:

coverage is numeric (int or float)
coverage is strictly between 0 and 1

If invalid values are provided, a clear ValueError is raised.

Scope

This PR focuses only on input validation and safe handling of the parameter.
It does not implement prediction interval logic, which can be handled
separately.

Why this helps- Improves robustness of the tool
Prepares the tool interface for probabilistic forecasting support
 Especially useful for LLM-driven workflows where input validation is critical

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,18 @@ python3 -m pip install -e .
 
 # With all optional dependencies
 python3 -m pip install -e ".[all]"
+### ⚠️ Windows + Python 3.14 note
+
+On Windows with Python 3.14, installing with:
+    pip install -e ".[all]"
+may fail when building `scipy` from source due to unsupported compiler flags (C17).
+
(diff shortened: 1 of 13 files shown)


Shashankss1205:
fit_predict_tool was split into fit + predict in the refactor, and predict already accepts and forwards a coverage parameter (tools/fit_predict.py). This branch also carries the unrelated router/profiling/tracing files from your other PRs and never registers anything in server.py. Closing as obsolete.

PR #300, opened 2026-04-25
DOC: fix broken documentation links

biru-codeastromer:
#### What does this implement/fix?

This PR keeps the docs fix intentionally small:

- Fixes broken architecture diagram paths so Sphinx can find the images under `docs/source/_static`.
- Removes the missing `ideal-mcp-tools` page from the docs toctree.
- Retargets the corresponding usage guide link to the existing API reference instead of the missing page.

#### Validation

```bash
git diff --check
.venv/bin/sphinx-build -E -b html docs/source docs/build/html
```

Both commands pass. The normal docs build still reports unrelated pre-existing warnings in other pages, but this PR removes the missing-page and broken-image warnings without rewriting those sections.


Shashankss1205:
This was correct at the time. The docs were restructured in #435, which deleted architecture.md/usage-examples.md, dropped the ideal-mcp-tools toctree entry, and moved the diagrams to developer/architecture.md with working _static paths. All three broken links this PR targeted are already gone from main. Closing as resolved.

Issue #287, opened 2026-04-23
MCP usability gaps for agent-driven pipeline construction


ParminderSinghGithub:
## Summary  
While experimenting with an LLM-driven workflow using sktime-mcp, I noticed a few usability gaps. Although it’s possible to build an end-to-end forecasting pipeline using the available tools, the process feels more cumbersome than it needs to be—especially for agent-based workflows.

## Description  
I walked through a simple pipeline construction using only the MCP tools (discover → select → validate → instantiate → run). The workflow does complete successfully, but it requires extra steps and some trial-and-error that could likely be avoided with a few small improvements to the tool outputs and metadata.

### 1. Estimator discovery lacks descriptive context  
The `list_estimators` tool currently returns only basic metadata (`name`, `task`, `module`, `tags`) without any explanation of what each estimator actually does.

**Impact:**
- An agent has to call `describe_estimator` repeatedly just to understand candidate options  
- This adds unnecessary overhead during the discovery phase  

---

### 2. Composability metadata is not exposed  
Although the validator clearly understands which components are compatible with each other, there’s no way to query this composability information directly through an MCP tool.

**Impact:**
- Agents are forced to rely on trial-and-error via `validate_pipeline`  
- Pipeline construction becomes iterative rather than guided and predictable  

---

### 3. Data loading configuration is not clearly defined  
The `load_data_source` tool accepts a generic `config` object, but the expected structure varies by source type (CSV, SQL, etc.) and isn’t explicitly documented.

**Impact:**
- Agents have to infer valid configurations from examples  
- This introduces ambiguity and makes programmatic pipeline construction less reliable  

---

## Why this matters  
sktime-mcp is clearly designed with agentic workflows in mind, where an LLM assembles pipelines autonomously. While the current system is functionally complete, these gaps introduce avoidable friction—leading to more tool calls and a heavier reliance on implicit assumptions.

## Suggested direction  
A few targeted enhancements could make a meaningful difference:

- Add a short description field to `list_estimators`  
- Expose composability metadata (e.g., compatible predecessors and successors) via a dedicated MCP tool  
- Provide clearer, structured schemas for `load_data_source` configurations based on source type  

These changes would make the system more intuitive and better aligned with how LLM agents actually operate.

## Context  
This feedback comes from simulating a full end-to-end agent workflow using the available MCP tools (discovery → selection → validation → instantiation).

Happy to explore any of these ideas further if they fit with the intended direction.

sArThAk1612:
[...]

PR #303, opened 2026-04-25
FEAT: add descriptions to list_estimators results

biru-codeastromer:
#### Reference Issues/PRs

Addresses one part of #287.

Adds a compact `description` field to estimator summaries returned by `list_estimators`.

The description is derived from the existing estimator docstring already loaded into `EstimatorNode`, normalized to a single line, and capped at 200 characters. This gives agents useful context during estimator discovery without requiring repeated `describe_estimator` calls for every candidate.

Also updates the README return description and adds focused tests for the new summary field.

No new dependencies.

- Whether 200 characters is the right length for `list_estimators` summaries.
- Whether the field name `description` is preferred over `summary`.
- Whether the description should be empty for estimators without docstrings, as implemented here.

#### Any other comments?

This intentionally addresses only the estimator-discovery context gap from #287. The other ideas in that issue, such as composability metadata and data-source schemas, are left for separate PRs.

##### For all contributions
- [ ] I have added myself to the list of contributors. I can do this separately if maintainers want it for sktime-mcp contributions.
- [ ] Optionally, I have updated CODEOWNERS. Not applicable.
- [x] I have added unit tests and made sure they pass locally.

##### For new estimators
- [ ] Not applicable; this does not add a new estimator.

#### Validation

```bash
.venv/bin/python -m ruff check src/sktime_mcp/registry/interface.py src/sktime_mcp/tools/list_estimators.py tests/test_core.py
.venv/bin/python -m ruff format --check src/sktime_mcp/registry/interface.py src/sktime_mcp/tools/list_estimators.py tests/test_core.py
git diff --check
.venv/bin/python -m pytest tests/test_core.py -q
.venv/bin/python -m pytest -q
```

All commands pass. The full test suite reports two pre-existing async warnings unrelated to this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ Discover estimators by task type and capability tags.
 }
 ```
 
-**Returns:** List of matching estimators with name, task, and summary info.
+**Returns:** List of matching estimators with name, task, short description, and summary info.
 
 ---
 
--- a/src/sktime_mcp/registry/interface.py
+++ b/src/sktime_mcp/registry/interface.py
@@ -39,6 +39,17 @@ class EstimatorNode:
     hyperparameters: dict[str, Any] = field(default_factory=dict)
     docstring: Optional[str] = None
 
+    def _description(self, max_length: int = 200) -> str:
+        """Return a compact, single-line description for list responses."""
+        if not self.docstring:
+            return ""
+
+        description = " ".join(self.docstring.split())
+        if len(description) <= max_length:
+            return description
+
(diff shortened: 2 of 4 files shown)


Shashankss1205:
This PR is written against list_estimators, which was replaced by query_registry with a different summary shape (registry/interface.py), so it cannot be rebased as-is. The idea of a one-line description per result is reasonable, but the audit already flags query_registry payloads as too large, so it needs a maintainer decision first. Closing as obsolete.

PR #304, opened 2026-04-25
FEAT: expose composition metadata as an MCP tool

biru-codeastromer:
## Reference Issues/PRs

Addresses one part of #287.

This PR exposes the existing composition validator metadata through a new MCP tool, `get_valid_compositions`. Given an estimator name, the tool returns the task types that can validly precede or follow that estimator in a pipeline.

The implementation is intentionally small: it reuses `CompositionValidator.get_valid_compositions()` instead of adding new composition rules, registers the tool in the MCP server, and adds tests for the direct tool function and server-level tool dispatch.

No.

- Whether `get_valid_compositions` is the right MCP tool name
- Whether the response shape is useful for agents building sktime pipelines
- Whether returning task-type compatibility is the right first step before adding estimator-level candidate suggestions

## Did you add any tests for the change?

Yes. The PR adds coverage for:

- successful composition metadata lookup for a known estimator
- structured error output for an unknown estimator
- MCP server registration and dispatch for the new tool

## Any other comments?

Validation run locally:

```
.venv/bin/python -m ruff check src/sktime_mcp/server.py src/sktime_mcp/tools/composition.py tests/test_core.py
.venv/bin/python -m ruff format --check src/sktime_mcp/server.py src/sktime_mcp/tools/composition.py tests/test_core.py
git diff --check
.venv/bin/python -m pytest tests/test_core.py -q
.venv/bin/python -m pytest -q
.venv/bin/sphinx-build -E -b html docs/source docs/build/html
```

Full tests passed with two pre-existing async warnings. The docs build completed successfully with pre-existing documentation warnings unrelated to this change.

## Checklist

- [x] Added tests for the change
- [x] Ran local tests
- [x] No new dependency introduced

### DIFF
--- a/docs/source/user-guide.md
+++ b/docs/source/user-guide.md
@@ -66,6 +66,7 @@ The `sktime-mcp` server exposes a suite of tools that your AI assistant uses on 
 |----------------------|-------------------------------|----------------|
 | **Find models** | The assistant searches the sktime registry by task, tags, or keywords. | *"What forecasting models are available?"* |
 | **Create a model or pipeline** | An estimator or multi-step pipeline is instantiated with your chosen parameters. | *"Set up an ARIMA(1,1,1) model"* |
+| **Inspect valid pipeline steps** | The assistant checks what types of estimators can come before or after a component. | *"What can I compose after Detrender?"* |
 | **Run a forecast** | The model is fitted on your data and predictions are generated. | *"Forecast the airline dataset 12 months ahead"* |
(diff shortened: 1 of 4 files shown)


Shashankss1205:
The thing it wraps no longer exists: CompositionValidator was deleted in #507 because instantiate now delegates composition validation to sktime's craft(), which raises on invalid pipelines. There is no metadata source left to expose. Closing as obsolete.

Issue #214, opened 2026-04-15
[BUG]  `main()` writes startup message to stdout which may corrupt MCP stdio stream


Om-Dhondge:
# [BUG] `main()` writes startup message to stdout may corrupt MCP stdio stream

## Summary

`server.py` prints a startup message to stdout before the MCP server initialises. Since the MCP stdio transport uses stdout as its JSON-RPC channel, any non-JSON bytes written there before the protocol handshake could cause MCP clients to fail parsing the stream.

## Steps to Reproduce

```bash
pip install -e .

# Capture what a connecting MCP client would see on stdout:
python -c "from sktime_mcp.server import main; main()" > captured_stdout.txt &
sleep 1 && kill %1

cat captured_stdout.txt
# Starting sktime-mcp server...
```

The first bytes on stdout are plain text, not a JSON-RPC frame.

## Root Cause

[`server.py`, line 787](https://github.com/sktime/sktime-mcp/blob/main/src/sktime_mcp/server.py#L787):

```python
def main():
    """Main entry point."""
    print("Starting sktime-mcp server...")   # writes to stdout
    asyncio.run(run_server())
```

`print()` defaults to `sys.stdout`. The MCP stdio transport reads from `sys.stdin` and writes to `sys.stdout` using newline-delimited JSON-RPC. A plain-text line on stdout before the protocol starts is not valid JSON-RPC and will be misinterpreted by strict clients.

## Impact

Behaviour depends on the MCP client implementation:
- **Lenient clients** (e.g. those that silently skip non-JSON lines) may be unaffected.
- **Strict clients** that treat any non-JSON line as a protocol error will fail to connect.

The project already guards against stderr noise via issue #194. Stdout is a more sensitive channel as it carries the actual protocol and not just diagnostics.

## Possible Fix

`logger` is already defined in `server.py` and routes to stderr, which is the appropriate channel for startup messages:

```python
def main():
    """Main entry point."""
    logger.info("Starting sktime-mcp server...")   # stderr — invisible to MCP client
    asyncio.run(run_server())
```

This would bring `main()` in line with how the rest of the server already handles diagnostics.

## Environment

- `sktime-mcp` version: `0.1.0`
- Python: `>=3.10`
- MCP SDK: `mcp>=1.0.0`

Happy to submit a PR for this, the change is a one-liner:

```python
logger.info("Starting sktime-mcp server...")
```

PR #309, opened 2026-04-26
Add MCP protocol test suite — stream cleanliness, tool registration, error contracts, JSON serializability

### LINKED ISSUE #310: MCP protocol layer has no test coverage — stdout corruption, silent tool loss, and serialization failures go undetected
### What's missing

sktime-mcp has tests for tool logic — `test_core.py`, `test_evaluate.py`, etc. — but nothing that tests whether the server works correctly *as an MCP server*. These are different failure surfaces, and right now the protocol layer is completely uncovered.

Concretely, none of the following are currently verified by any test:

- **Stdout cleanliness** — anything printed to stdout during server startup silently corrupts the JSON-RPC transport. An LLM client connecting over stdio receives garbled data and the connection fails. This happened in #214. The fix was merged but there's nothing preventing it from being reintroduced.

- **Tool registration** — a broken `@mcp.tool()` decorator or a silent import error produces no exception. The server starts, the tool list is just empty. An LLM connecting to the server has no tools and no indication why.

- **Tool descriptions** — LLMs select tools by reading the description field. A tool with no description is invisible to the agent. This can happen when a tool is added without a docstring or when one is accidentally removed during refactoring. No test catches it today.

- **Structured error responses** — MCP clients expect errors in the result payload (`{"success": false, "error": "..."}`). If a tool raises an unhandled Python exception instead, the MCP transport crashes and the LLM gets a protocol error rather than a usable message. This is never tested.

- **JSON serializability** — tools that return NumPy or Pandas types break the serialization pipeline. `sanitize_for_json` is supposed to catch these but the output is never actually round-tripped through `json.dumps()`. This was the root cause of #197 and can silently recur.

### What a fix looks like

A `test_mcp_protocol.py` that uses `create_connected_server_and_client_session` from the MCP SDK to spin up an in-memory client/server pair and exercises these properties through the protocol, the same way a real LLM client would.

The test file should not call internal server methods directly — those don't reflect the actual protocol contract and several don't work the way you'd expect (e.g. `server.list_tools()` is not a coroutine).

### Why this matters

Tool logic tests passing while the protocol layer is broken is a real scenario. If `test_mcp_protocol` goes red, it means the server is broken for all LLM clients regardless of whether individual tool tests pass. That distinction is worth making visible in CI as a separate job.

nXtCyberNet:
## Add MCP protocol test suite

sktime-mcp had solid tests for what the tools *do* — but nothing that tested whether they work *as an MCP server*. That's a different failure surface entirely, and it had been quietly invisible.

Two bugs made this concrete. Issue #214 was a stdout corruption: something was printing during server startup, which silently broke the JSON-RPC transport for every LLM client connecting over stdio. Issue #197 was a serialization gap: `sanitize_for_json` wasn't catching all NumPy and Pandas types, so tools could return objects that weren't actually JSON-serializable. Both bugs were fixed — but there was nothing stopping them from coming back.

This PR makes those fixes permanent, and closes the broader gap at the protocol layer.

---

### What's in here

**`tests/test_mcp_protocol.py`** — 7 tests that exercise the MCP contract directly, using an in-memory client/server session via `create_connected_server_and_client_session`. Every test talks to the server the same way a real LLM client would: over the protocol, not by calling internal methods.

The tests cover:

- **Stream cleanliness** — redirects both stdout and stderr before the session even opens, so lazy-loading side effects during startup are caught too. Regression lock for #214.
- **Tool registration** — asserts the tool list is actually non-empty. A broken `@mcp.tool()` decorator or a silent import error produces no exception, just an empty list. An LLM connecting to the server would have no tools and no idea why.
- **Tool descriptions** — every registered tool must have a non-empty description. LLMs select tools purely by description; a missing one makes that tool invisible to the agent with zero noise anywhere in the logs.
- **Structured error responses** — uses a guaranteed-bad input (a nonsense estimator name) and asserts `success: false` unconditionally. The failure path is always exercised, never conditionally skipped.
- **JSON serializability** — parses the response, then re-serializes the Python object with `json.dumps()`. This is what catches NumPy/Pandas types that survived `sanitize_for_json`. Covers #197.
- **Concurrent handle uniqueness** — two parallel instantiation calls must produce distinct handles. Catches state collisions under concurrent load.
- **Minimal tool contracts** — verifies that core tools return the keys LLM agents depend on for reasoning.

**`.github/workflows/ci.yml`** — two additions:
- Coverage reporting on the main test job (uploads on Python 3.12 only to avoid triple-counting)
- A separate `test-mcp-protocol` job so protocol failures show up as their own red check, distinct from tool logic failures

---

### Implementation note

The tests went through several iterations. The first pass called `server.list_tools()` and `server.call_tool()` directly on the server object — those aren't coroutines and don't accept arguments that way. The second pass used a shared async pytest fixture to hold the client session — anyio cancel scopes can't span task boundaries, so teardown crashed with `RuntimeError: Attempted to exit cancel scope in a different task`. The final approach gives each test its own `async with` block, which keeps the context manager's enter and exit in the same task.

---

All 7 tests pass on Python 3.12:

```
collected 7 items
test_transport_stream_cleanliness PASSED
test_tool_list_not_empty PASSED
test_all_tools_have_descriptions PASSED
test_tool_error_response_is_structured PASSED
test_list_estimators_returns_valid_json PASSED
test_concurrency_handle_uniqueness PASSED
test_minimal_tool_contracts PASSED
7 passed in 3.23s
```

Closes #310 #214 . Catches the class of bug in #197 automatically going forward.


Shashankss1205:
This is the only PR in the backlog that adds protocol-level coverage (in-memory client/server session, stdout cleanliness, JSON re-serialisation) and it still fits current main. To make it mergeable: (1) rebase; the only conflict is the dev extras block in pyproject.toml (keep pytest-cov, black is gone); (2) update tool names: list_estimators -> query_registry (response key is results, not estimators), instantiate_estimator {estimator: X} -> instantiate {spec: 'NaiveForecaster()'}; I verified all 7 tests pass with exactly those changes; (3) drop mcp_test.yaml, which duplicates the ci.yml matrix, and let the file run under the existing pytest step.

Shashankss1205:
This is the only PR in the backlog that adds protocol-level coverage (in-memory client/server session, stdout cleanliness, JSON re-serialisation) and it still fits current main. To make it mergeable: (1) rebase; the only conflict is the dev extras block in pyproject.toml (keep pytest-cov, black is gone); (2) update tool names: list_estimators -> query_registry (response key is results, not estimators), instantiate_estimator {estimator: X} -> instantiate {spec: 'NaiveForecaster()'}; I verified all 7 tests pass with exactly those changes; (3) drop mcp_test.yaml, which duplicates the ci.yml matrix, and let the file run under the existing pytest step.

Shashankss1205:
This is the only PR in the backlog that adds protocol-level coverage (in-memory client/server session, stdout cleanliness, JSON re-serialisation) and it still fits current main. To make it mergeable: (1) rebase; the only conflict is the dev extras block in pyproject.toml (keep pytest-cov, black is gone); (2) update tool names: list_estimators -> query_registry (response key is results, not estimators), instantiate_estimator {estimator: X} -> instantiate {spec: 'NaiveForecaster()'}; I verified all 7 tests pass with exactly those changes; (3) drop mcp_test.yaml, which duplicates the ci.yml matrix, and let the file run under the existing pytest step.

Me-Priyank:
Really glad someone's doing protocol-level coverage here.

Tried running it, the tool names have moved since April: `list_estimators` → `query_registry` (returns `results`, not `estimators`) and `instantiate_estimator` → `instantiate` (takes `spec`, a craft string). They now fall through to the unknown-tool branch, which has no `success` key, so `validate_mcp_response` trips at line 110. `pyproject.toml` also needs a rebase, `black` was dropped from the dev extras in `b78ca00`.

One suggestion: the `assert f_err.getvalue() == ""` might be worth dropping rather than fixing. The stdio spec says servers **MAY** write to stderr and clients shouldn't read it as an error signal, and the logging handler binds `sys.stderr` at import (`server.py:104`), so `redirect_stderr` never sees it anyway.

PR #323, opened 2026-04-27
FIX: reject invalid cv_folds values in evaluate_estimator_tool

### LINKED ISSUE #322: [BUG] evaluate_estimator_tool silently coerces zero or negative cv_folds to 1
## Problem

`evaluate_estimator_tool` currently accepts `cv_folds=0` and negative values, but instead of rejecting them it silently coerces them to a single-fold evaluation and returns `success: true`. For MCP/LLM workflows, this is misleading because the caller asked for an invalid number of folds and receives a valid-looking result.

## Reproduction

```python
from sktime_mcp.tools.evaluate import evaluate_estimator_tool

inst = instantiate_estimator_tool("NaiveForecaster", {"strategy": "last"})
handle = inst["handle"]

print(evaluate_estimator_tool(handle, "airline", 0))
print(evaluate_estimator_tool(handle, "airline", -1))
```

Observed behavior:

- `cv_folds=0` returns `success: true` with `cv_folds_run: 1`
- `cv_folds=-1` returns `success: true` with `cv_folds_run: 1`

The response still reports the original invalid request in `cv_folds_requested`, which makes the result internally inconsistent.

## Root cause

In `src/sktime_mcp/tools/evaluate.py`, the tool currently does:

```python
folds = max(1, min(int(cv_folds), max(1, n - 1)))
```

This clamps invalid non-positive inputs up to `1` instead of rejecting them.

## Expected behavior

Invalid fold counts should fail fast with a clear validation error, for example:

```python
{"success": False, "error": "cv_folds must be a positive integer."}
```

## Why this matters

This is an agent-trust issue. An LLM can incorrectly believe it successfully ran 0-fold or negative-fold cross-validation, when the tool actually executed a different evaluation than requested. Explicit validation would make failures easier to recover from and keep evaluation outputs aligned with user intent.

## Suggested fix

Before computing the splitter:

- validate that `cv_folds` is an integer greater than 0
- return a clear error for `0`, negative values, or non-integer-like inputs
- add tests covering these invalid cases

biru-codeastromer:
## Reference Issues/PRs

Closes #322.

This PR makes `evaluate_estimator_tool` fail fast when `cv_folds` is invalid instead of silently coercing the request to a one-fold evaluation.

Before this change, calls like `cv_folds=0` or `cv_folds=-1` returned `success: true` and ran one fold anyway. That made the response inconsistent with the user request and could mislead MCP clients or LLM agents into believing the requested evaluation actually happened.

The new behavior rejects invalid values with:

```python
{"success": False, "error": "cv_folds must be a positive integer."}
```

Valid inputs still behave the same as before, including clamping large fold counts to the dataset size limit.

No.

- Whether the validation boundary for `cv_folds` is appropriate
- Whether accepting integer-like string inputs via `int(cv_folds)` is acceptable for tool-level robustness
- Whether the error wording should align with other parameter-validation responses

## Did you add any tests for the change?

Yes. The PR adds tests covering:

- `cv_folds=0`
- `cv_folds=-1`
- `cv_folds="abc"`
- `cv_folds=True`

All of these now return a clear validation error.

## Any other comments?

Validation run locally:

```
.venv/bin/python -m ruff check src/sktime_mcp/tools/evaluate.py tests/test_evaluate.py
.venv/bin/python -m ruff format --check src/sktime_mcp/tools/evaluate.py tests/test_evaluate.py
git diff --check
.venv/bin/python -m pytest tests/test_evaluate.py -q
.venv/bin/python -m pytest -q
```

Full test suite passed locally with pre-existing async warnings unrelated to this change.

## Checklist

- [x] Added tests for the change
- [x] Ran local tests
- [x] No new dependency introduced

### DIFF
--- a/src/sktime_mcp/tools/evaluate.py
+++ b/src/sktime_mcp/tools/evaluate.py
@@ -49,9 +49,20 @@ def evaluate_estimator_tool(
     y = data_result["data"]
     X = data_result.get("exog")
 
+    if isinstance(cv_folds, bool):
+        return {"success": False, "error": "cv_folds must be a positive integer."}
+
+    try:
+        requested_folds = int(cv_folds)
+    except (TypeError, ValueError):
+        return {"success": False, "error": "cv_folds must be a positive integer."}
+
+    if requested_folds < 1:
+        return {"success": False, "error": "cv_folds must be a positive integer."}
+
     try:
         n = len(y)
-        folds = max(1, min(int(cv_folds), max(1, n - 1)))
+        folds = min(requested_folds, max(1, n - 1))
         # Exactly `folds` backtest windows: train grows, last fold uses n-1 obs before last point.
         initial_window = max(1, n - folds)
         cv = ExpandingWindowSplitter(initial_window=initial_window, step_length=1, fh=[1])
@@ -69,7 +80,7 @@ def evaluate_estimator_tool(
             "success": True,
             "results": metrics,
             "cv_folds_run": len(metrics),
-            "cv_folds_requested": int(cv_folds),
+            "cv_folds_requested": requested_folds,
         }
     except Exception as e:
         logger.exception("Error during evaluate")
--- a/tests/test_evaluate.py
+++ b/tests/test_evaluate.py
@@ -38,5 +38,25 @@ def test_evaluate_estimator_tool():
         executor._handle_manager.release_handle(handle)
 
 
+@pytest.mark.parametrize("cv_folds", [0, -1, "abc", True])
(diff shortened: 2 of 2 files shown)


Shashankss1205:
Fixed independently in #527 (91d46b1): evaluate now returns 'cv_folds must be a positive integer, got N' for zero/negative values instead of clamping, and non-integer values are rejected by schema validation before the tool runs. The evaluate_estimator_tool this PR patches was replaced by evaluate in #505. Closing along with #322 as fixed.

PR #340, opened 2026-04-28
FIX: validate top-level argument types in instantiation tools

### LINKED ISSUE #339: [BUG] instantiation tools crash or mis-handle invalid top-level argument types
## Summary
The instantiation tools validate `params`, but do not validate the top-level `estimator` and `components` arguments before using them. That leads to Python exceptions or misleading behavior instead of structured MCP errors.

## Reproduction
Using current `main`:
```python
from sktime_mcp.tools.instantiate import instantiate_estimator_tool, instantiate_pipeline_tool

instantiate_estimator_tool(["ARIMA"])
# TypeError: unhashable type: list

instantiate_pipeline_tool("ARIMA")
# treated as ["A", "R", "I", "M", "A"] and returns misleading validation errors

instantiate_pipeline_tool(123)
# TypeError: object of type int has no len()
```

## Expected behavior
- `instantiate_estimator_tool` should reject non-string estimator names with `success: false`
- `instantiate_pipeline_tool` should require a non-empty list of strings and reject malformed inputs with `success: false`

## Why this matters
These are core MCP entrypoints. LLM clients can send malformed top-level argument types, and tool-level crashes are much harder to recover from than structured validation errors.

## Proposed fix
- validate that `estimator` is a non-empty string
- validate that `components` is a non-empty list of non-empty strings
- add focused tests for malformed top-level inputs

biru-codeastromer:
## Summary
- validate that `instantiate_estimator_tool` receives a non-empty string estimator name
- validate that `instantiate_pipeline_tool` receives a non-empty list of non-empty string component names
- add focused tests for malformed top-level inputs

## Why
Before this change, malformed top-level inputs could crash the tool or produce misleading behavior:
- `instantiate_estimator_tool(["ARIMA"])` raised `TypeError`
- `instantiate_pipeline_tool("ARIMA")` treated the string as character-by-character components
- `instantiate_pipeline_tool(123)` raised `TypeError`

For MCP clients and LLM agents, structured validation errors are much easier to recover from than Python exceptions or confusing fallback behavior.

Closes #339

## Testing
- `.venv/bin/python -m ruff check src/sktime_mcp/tools/instantiate.py tests/test_param_validation.py`
- `.venv/bin/python -m ruff format --check src/sktime_mcp/tools/instantiate.py tests/test_param_validation.py`
- `git diff --check`
- `.venv/bin/python -m pytest tests/test_param_validation.py -q`
- `.venv/bin/python -m pytest -q`

### DIFF
--- a/src/sktime_mcp/tools/instantiate.py
+++ b/src/sktime_mcp/tools/instantiate.py
@@ -99,6 +99,55 @@ def _validate_params(
     return {"valid": True, "warnings": warnings}
 
 
+def _validate_estimator_name(estimator: Any) -> dict[str, Any]:
+    """Validate the top-level estimator argument."""
+    if not isinstance(estimator, str):
+        return {
+            "valid": False,
+            "error": (
+                f"'estimator' must be a non-empty string, got {type(estimator).__name__}. "
+                'Example: "ARIMA"'
+            ),
+        }
+
+    if not estimator.strip():
+        return {
+            "valid": False,
+            "error": "'estimator' must be a non-empty string.",
+        }
+
+    return {"valid": True}
+
+
+def _validate_components(components: Any) -> dict[str, Any]:
+    """Validate the top-level pipeline components argument."""
+    if not isinstance(components, list):
+        return {
+            "valid": False,
+            "error": (
+                f"'components' must be a non-empty list of estimator names, "
+                f"got {type(components).__name__}."
+            ),
+        }
+
+    if not components:
+        return {
+            "valid": False,
+            "error": "Pipeline cannot be empty",
+        }
+
+    for i, component in enumerate(components):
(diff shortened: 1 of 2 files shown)


Shashankss1205:
The two tools this validates were merged into a single instantiate(spec) tool in #508, and that tool already rejects empty/non-string specs with "A valid 'spec' string is required."; the MCP layer additionally enforces the string type from the schema. Closing along with #339 as fixed.

Shashankss1205:
The two tools this validates were merged into a single instantiate(spec) tool in #508, and that tool already rejects empty/non-string specs with "A valid 'spec' string is required."; the MCP layer additionally enforces the string type from the schema. Closing along with #339 as fixed.

PR #347, opened 2026-04-28
fix: map NaN/Inf to null in sanitize_for_json

SarthakB11:
Follow-up to #198 (merged today).

## Bug

The merged `sanitize_for_json` returns `float(obj)` for `np.floating`, which produces `nan`/`inf` when the input is `np.float64('nan')` or `np.float64('inf')`. Strict JSON has no representation for these:

```python
>>> import json, numpy as np
>>> from sktime_mcp.server import sanitize_for_json
>>> json.dumps({'x': sanitize_for_json(np.float64('nan'))}, allow_nan=False)
ValueError: Out of range float values are not JSON compliant
```

Even with default `allow_nan=True`, the output `{"x": NaN}` is **invalid JSON** per RFC 8259 — strict consumers (most JSON-RPC parsers, including the one MCP uses over stdio) reject it.

## When this bites

Real cases that produce NaN/Inf in tool responses:
- All-NaN input columns → NaN means/stds in describe-style outputs
- Empty CV folds → `inf` MASE
- Division-by-zero in metric computations

## Fix

Map NaN and Inf to `None` (JSON `null`) inside the `np.floating` branch. Five-line change.

## Tests

Strengthens the existing `test_nan` (which previously only verified `json.dumps(..., default=str)` worked — i.e. the escape hatch, not the contract). Adds `test_inf` and `test_nan_in_dict` for realistic usage.

### DIFF
--- a/src/sktime_mcp/server.py
+++ b/src/sktime_mcp/server.py
@@ -95,7 +95,11 @@ def sanitize_for_json(obj):
         if isinstance(obj, np.integer):
             return int(obj)
         if isinstance(obj, np.floating):
-            return float(obj)
+            f = float(obj)
+            # JSON has no representation for NaN/Inf — map to null.
+            if f != f or f == float("inf") or f == float("-inf"):
+                return None
+            return f
         if isinstance(obj, np.bool_):
             return bool(obj)
         if isinstance(obj, np.ndarray):
--- a/tests/test_sanitize.py
+++ b/tests/test_sanitize.py
@@ -135,5 +135,19 @@ class TestEdgeCases:
         assert sanitize_for_json({}) == {}
 
     def test_nan(self):
-        result = sanitize_for_json(np.float64("nan"))
-        json.dumps(result, default=str)
+        # NaN/Inf have no JSON representation; map to null so json.dumps
+        # works without the default=str escape hatch.
+        assert sanitize_for_json(np.float64("nan")) is None
+        json.dumps(sanitize_for_json(np.float64("nan")))
+
+    def test_inf(self):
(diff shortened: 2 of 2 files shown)


SarthakB11:
Bumping this, still mergeable with no conflicts. Small, isolated fix (RFC 8259 compliance for NaN/Inf) with tests. If it helps, I have two other small open PRs against this same repo (#114, #515) that don't overlap with this one code-wise, happy for a batch look if that's easier than three separate passes.

Shashankss1205:
This is the right place for the fix and it still applies cleanly to current main (311 tests pass). One gap: the check only runs in the np.floating branch, but predict returns builtin Python floats (from Series.to_dict()), so the audit scenario (NaiveForecaster with sp=6 on 4 points, predict(horizon=6)) still emits bare NaN on the wire. Please also handle builtin float (math.isfinite in the already-JSON-safe branch, or one shared helper), add a test with float('nan')/float('inf') inside a dict, and ideally switch the json.dumps call in server.py to allow_nan=False as a guard. With that, we merge.

Me-Priyank:
Had a look at this, the check only fires in the `np.floating` branch, but the paths that actually carry forecasts hand over builtin floats: `Series` via `tolist()` at `server.py:196`, `DataFrame` via `to_dict()` at `:198`, and `predict` skips the sanitiser entirely (`executor.py:645-661`).

With the patch applied I still get `sanitize_for_json(pd.Series([1.0, np.nan, 3.0]))` → `[1.0, NaN, 3.0]`. Extending to the scalar branch at `:207` would cover those. The test hunk also needs a rebase past #515.

PR #349, opened 2026-04-28
FIX: reject boolean horizon values in forecasting tools

### LINKED ISSUE #348: [BUG] forecasting tools accept boolean horizon values and coerce True to 1

`fit_predict_tool` and `fit_predict_async_tool` currently accept boolean values for `horizon` because the validation uses `isinstance(horizon, int)`, and in Python `bool` is a subclass of `int`.

That means:

```python
fit_predict_tool(handle, "airline", horizon=True)
```

returns success and behaves like `horizon=1`, while:

```python
fit_predict_async_tool(handle, dataset="airline", horizon=True)
```

starts a background job with `horizon: True`.

#### Why this matters

For an MCP tool, `true`/`false` in JSON should not silently become forecast horizons. This is especially risky for LLM-driven tool use, where type slips can happen and a successful response hides the problem.

#### Steps to reproduce

```python
from sktime_mcp.tools.fit_predict import fit_predict_tool

result = instantiate_estimator_tool("NaiveForecaster")
handle = result["handle"]

fit_predict_tool(handle, "airline", horizon=True)
# currently succeeds and returns a 1-step forecast
```

#### Expected behavior

All forecasting tool entrypoints should reject boolean horizons with the same validation error used for other non-integer values.

#### Proposed fix

Treat `bool` as invalid in the shared horizon validator, and add regression tests for:
- `predict_tool(..., horizon=True/False)`
- `fit_predict_tool(..., horizon=True)`
- `fit_predict_async_tool(..., horizon=True)`

biru-codeastromer:
Closes #348

## Summary
- reject `True`/`False` in the shared horizon validator used by forecasting tools
- keep error messaging consistent with other non-integer horizon inputs
- add focused regression tests for sync, async, and predict entrypoints

## Problem
`fit_predict_tool` and `fit_predict_async_tool` accepted boolean horizon values because Python treats `bool` as a subclass of `int`.

Before this change:

```python
fit_predict_tool(handle, "airline", horizon=True)
```

returned success and behaved like a 1-step forecast.

Similarly:

```python
fit_predict_async_tool(handle, dataset="airline", horizon=True)
```

scheduled a background job with `horizon=True`.

For MCP/JSON tool use, that is a bad contract: `true`/`false` should not silently become valid forecast horizons.

## Fix
The shared `_validate_horizon()` helper now treats `bool` as invalid, so all forecasting entrypoints reject it consistently.

## Tests
Added regression coverage for:
- `predict_tool(..., horizon=True)`
- `predict_tool(..., horizon=False)`
- `fit_predict_tool(..., horizon=True)`
- `fit_predict_async_tool(..., horizon=True)`

## Verification
- `ruff check .`
- `ruff format --check .`
- `pytest -q`

### DIFF
--- a/src/sktime_mcp/tools/fit_predict.py
+++ b/src/sktime_mcp/tools/fit_predict.py
@@ -19,7 +19,7 @@ def _validate_horizon(horizon: int) -> dict[str, Any]:
     Checks if the horizon parameter is greater than 0 or not
     """
     warnings = []
-    if not isinstance(horizon, int):
+    if isinstance(horizon, bool) or not isinstance(horizon, int):
         return {
             "valid": False,
             "error": (
--- a/tests/test_async_custom_data.py
+++ b/tests/test_async_custom_data.py
@@ -186,6 +186,40 @@ class TestSyncCustomData:
             == "Either 'dataset' (e.g. 'airline') or 'data_handle' (from load_data_source) is required."
         )
 
+    def test_sync_rejects_boolean_horizon(self):
+        """Boolean horizon values should be rejected before execution."""
+        from sktime_mcp.tools.fit_predict import fit_predict_tool
+
+        result = fit_predict_tool(
+            estimator_handle="fake_handle",
+            dataset="airline",
+            horizon=True,
+        )
+
+        assert not result["success"]
+        assert (
+            result["error"] == "'horizon' must be an integer, got bool. Example: {\"horizon\": 12}"
+        )
+
+
+class TestAsyncHorizonValidation:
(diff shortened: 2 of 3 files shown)


Shashankss1205:
The one-line fix to _validate_horizon is correct for direct Python calls, but MCP clients are already protected: horizon is declared integer in the schema and the SDK rejects true before the tool runs. The branch also carries tests/test_async_custom_data.py, which main deleted and which targets the removed fit_predict tools. Closing along with #348 as not reachable over MCP; if you still want the one-liner as defence-in-depth, a fresh PR with just that line and a test is welcome.

PR #351, opened 2026-04-28
[ENH] Add structured config schemas and validation for load_data_source

### LINKED ISSUE #350: [ENH] Add structured config schemas for load_data_source by source type
As identified in #287, the `load_data_source` tool accepts a generic 
config object but the expected structure varies by source type (CSV, 
SQL, file, etc.) and is not documented or validated. This causes agents 
to guess valid configurations, introducing ambiguity and unreliable 
behavior in agentic workflows.

This issue proposes adding explicit structured schemas per source type 
so agents receive clear, predictable config contracts.

Relates to #287.

mak4x13:
Closes #350. Relates to #287.

## What this does
Adds `data_schemas.py` with explicit per-source-type schemas (pandas, sql, file, url) 
and a `validate_config` function that checks required fields before hitting the executor.

## Why
Agents currently receive no feedback when passing an invalid or incomplete config, 
causing silent failures. This returns a structured error with missing fields and 
an example config so agents can self-correct.

### DIFF
--- a/src/sktime_mcp/tools/data_schemas.py
+++ b/src/sktime_mcp/tools/data_schemas.py
@@ -4,9 +4,9 @@ from typing import Any
 
 SCHEMAS = {
     "pandas": {
-        "required": ["type", "data", "time_column", "target_column"],
-        "optional": [],
-        "description": "Load from a pandas-compatible dict with date and value keys.",
+        "required": ["type", "data"],
+        "optional": ["time_column", "target_column", "exog_columns", "frequency"],
+        "description": "Load from a pandas-compatible dict or DataFrame.",
         "example": {
             "type": "pandas",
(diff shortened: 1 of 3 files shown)


mak4x13:
@Shashankss1205 PR updated based on Copilot review:

Aligned schemas with actual adapter contracts
Added conditional SQL validation
Added unit tests in tests/test_data_schemas.py
Fixed return shape to always include the validation field

Happy to iterate on any further feedback.

Shashankss1205:
Since it was opened, load_data_source gained a run_async flag (#468), so the PR conflicts in data_tools.py, and main now returns a structured {success: false, error: ...} for every malformed config (missing type, unknown type with the list of valid types, missing path/data/url/connection info) from the adapters themselves. A parallel SCHEMAS table would duplicate those checks and drift. Closing; schema-level per-type keys are tracked in #350.

PR #353, opened 2026-04-28
FIX: move fit_predict_tool to correct import in server.py

### LINKED ISSUE #352: [BUG] server.py imports fit_predict_tool from wrong module causing startup ImportError
## Problem
`server.py` imports `fit_predict_tool` from `data_tools` but the 
function is defined in `fit_predict.py`. This causes an `ImportError` 
at server startup.

## Reproduction
```python
from sktime_mcp.tools.data_tools import fit_predict_tool
# ImportError: cannot import name 'fit_predict_tool' from 'data_tools'
```

## Expected behavior
Server starts without ImportError.

## Fix
I have a fix ready and will open a PR immediately.

Sandrakimiring:
#### Reference Issues/PRs
Fixes #352

`server.py` imported `fit_predict_tool` from `sktime_mcp.tools.data_tools` 
but the function is defined in `sktime_mcp.tools.fit_predict`. This would 
cause an `ImportError` at server startup.

Fixed by removing it from the `data_tools` import block and adding it 
to the existing `fit_predict` import block.

No new dependencies.

The import change in `src/sktime_mcp/server.py` and the 3 new tests 
in `tests/test_core.py` in `TestServerImports`.

#### PR checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,23 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: 24.3.0
+    hooks:
+      - id: black
+        language_version: python3.10
+
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.3.4
+    hooks:
+      - id: ruff
+        args: [ --fix ]
+      - id: ruff-format
+
+  - repo: local
+    hooks:
+      - id: pytest
+        name: pytest
+        entry: pytest
+        language: system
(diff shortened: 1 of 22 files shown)


Sandrakimiring:
Note: this branch includes the previously merged commit #332. The only new change is the server.py import fix and TestServerImports in test_core.py. Happy to rebase if needed.

Shashankss1205:
How do you aim to fix the code by adding a test ?

PR #355, opened 2026-04-28
FIX: validate horizon parameter in fit_predict, predict, and fit_predict_async tools

### LINKED ISSUE #354: [BUG] horizon parameter not validated in fit_predict, predict, and fit_predict_async tools
## Problem
Three tools accept `horizon: int` but pass it through with no validation:

- `fit_predict_tool` — passes bad horizon to executor silently
- `predict_tool` — `range(1, horizon+1)` produces `fh=[]` when horizon≤0
- `fit_predict_async_tool` — creates a background job with a bad horizon

## Reproduction
```python
from sktime_mcp.tools.fit_predict import predict_tool
result = predict_tool("some_handle", horizon=0)
# No error — silently passes fh=[] to executor
```

## Expected behavior
Values less than 1 should return:
`{"success": False, "error": "horizon must be a positive integer."}`

## Fix
I have a fix ready and will open a PR immediately.

Sandrakimiring:
#### Reference Issues/PRs
Fixes #354

Three tools accepted `horizon: int` but passed it through with no validation:

- `fit_predict_tool` — passed bad horizon to executor silently
- `predict_tool` — `range(1, horizon+1)` produced `fh=[]` when horizon≤0
- `fit_predict_async_tool` — created a background job with a bad horizon

Added a guard as the first check in each function — before any executor,
asyncio, or job-manager code is reached. Values less than 1 now return:
`{"success": False, "error": "horizon must be a positive integer."}`

No new dependencies.

The 3 guards in `src/sktime_mcp/tools/fit_predict.py` and the 9 new 
tests in `TestHorizonValidation` in `tests/test_core.py`.

#### PR checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,23 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: 24.3.0
+    hooks:
+      - id: black
+        language_version: python3.10
+
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.3.4
+    hooks:
+      - id: ruff
+        args: [ --fix ]
+      - id: ruff-format
+
+  - repo: local
+    hooks:
+      - id: pytest
+        name: pytest
+        entry: pytest
+        language: system
+        pass_filenames: false
+        always_run: true
+        
--- a/Makefile
+++ b/Makefile
@@ -1,17 +1,17 @@
-.PHONY: check test lint format help format-fix
+.PHONY: check test lint format help format-fix install-hooks
 
 help:
 	@echo "Available commands:"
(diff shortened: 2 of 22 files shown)


Shashankss1205:
Already handled on main: predict validates horizon through the shared _validate_horizon helper in src/sktime_mcp/tools/fit_predict.py (rejects <=0 and non-integers with a structured error), and fit_predict/fit_predict_async were removed in #487, so the other two guards have no target. Closing along with #354 as fixed.

Shashankss1205:
Already handled on main: predict validates horizon through the shared _validate_horizon helper in src/sktime_mcp/tools/fit_predict.py (rejects <=0 and non-integers with a structured error), and fit_predict/fit_predict_async were removed in #487, so the other two guards have no target. Closing along with #354 as fixed.

PR #357, opened 2026-04-29
test: add regression coverage for server startup imports

aneek22112007-tech:
## Problem
A recent issue caused the MCP server to fail at startup due to a broken tool import (`cannot import name 'fit_predict_tool' from sktime_mcp.tools.data_tools`). This happens when tools are refactored, renamed, or moved, but the main `server.py` file is not updated to reflect those changes.

## Solution
This PR does two things:
1. Includes the fix for the broken `fit_predict_tool` import in `server.py`.
2. Adds `tests/test_server_startup.py` to act as a fast, isolated regression guard. 
   - `test_server_import_success`: verifies `import sktime_mcp.server` succeeds without any `ImportError`.
   - `test_server_tools_registration`: validates that the underlying MCP server object properly initializes and has critical expected tools registered (e.g., `fit_predict`, `load_data_source`).

## Testing
- Verified locally. All tests pass, including the new smoke tests.
- These tests are deterministic and run in milliseconds.

### DIFF
--- a/src/sktime_mcp/server.py
+++ b/src/sktime_mcp/server.py
@@ -18,7 +18,6 @@ from mcp.types import TextContent, Tool
 from sktime_mcp.composition.validator import get_composition_validator
 from sktime_mcp.tools.codegen import export_code_tool
 from sktime_mcp.tools.data_tools import (
-    fit_predict_tool,
     load_data_source_async_tool,
     load_data_source_tool,
     release_data_handle_tool,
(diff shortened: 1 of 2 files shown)


Shashankss1205:
The import this PR fixes disappeared with the removal of fit_predict_tool (#487), and the new test asserts the presence of list_estimators/instantiate_estimator/fit_predict, none of which exist after the tool consolidation. Several existing tests (tests/test_schema_types.py, tests/test_cli.py) already import sktime_mcp.server and enumerate list_tools(), so a broken import would fail CI today. Closing as obsolete.

PR #360, opened 2026-04-29
FIX: validate cancel_job delete flag type

### LINKED ISSUE #359: [BUG] cancel_job_tool treats non-boolean delete values as truthy and removes jobs unexpectedly
## Summary
`cancel_job_tool` does not validate the `delete` parameter type. Because Python treats non-empty strings as truthy, malformed MCP calls like `delete="false"` can still delete job records.

## Reproduced on current code (April 29, 2026)
```python
from sktime_mcp.runtime.jobs import get_job_manager, JobStatus
from sktime_mcp.tools.job_tools import cancel_job_tool

jm = get_job_manager()
job_id = jm.create_job("fit_predict", "handle", "ARIMA")
jm.update_job(job_id, status=JobStatus.COMPLETED)

cancel_job_tool(job_id, delete="false")
# -> {"success": True, "message": "Job '...' removed"}
```

The job record is removed even though the caller clearly did not provide a boolean.

## Why this matters
For MCP/JSON tools, `delete` should be a strict boolean. Silently treating strings or integers as truthy can cause destructive behavior from malformed tool calls.

## Expected behavior
`cancel_job_tool` should reject non-boolean `delete` values with a validation error instead of deleting anything.

## Suggested fix
Add a strict `isinstance(delete, bool)` guard near the start of `cancel_job_tool`, plus regression tests for string/integer inputs.

biru-codeastromer:
Closes #359

## Summary
- validate the `delete` flag type in `cancel_job_tool`
- reject non-boolean values before any job record is removed
- add a regression test for `delete="false"`

## Problem
`cancel_job_tool` treats the `delete` parameter as a plain truthy/falsy value.
That means malformed MCP calls like:

```python
cancel_job_tool(job_id, delete="false")
```

still delete job records, because non-empty strings are truthy in Python.

I reproduced this on current code by creating a completed job and calling `cancel_job_tool(job_id, delete="false")`, which returned success and removed the job.

## Fix
Add a strict boolean validation guard near the start of `cancel_job_tool`:
- if `delete` is not a real `bool`, return a validation error
- do not delete anything in that case

## Tests
Added regression coverage to ensure:
- `cancel_job_tool(job_id, delete="false")` returns an error
- the job record remains present after the invalid call

## Verification
- `ruff check .`
- `ruff format --check .`
- `pytest -q`

### DIFF
--- a/src/sktime_mcp/tools/job_tools.py
+++ b/src/sktime_mcp/tools/job_tools.py
@@ -100,6 +100,12 @@ def cancel_job_tool(job_id: str, delete: bool = False) -> dict[str, Any]:
     Returns:
         Dictionary with success status and message
     """
+    if not isinstance(delete, bool):
+        return {
+            "success": False,
+            "error": (f"Invalid delete type '{type(delete).__name__}'. Expected a boolean value."),
+        }
+
     job_manager = get_job_manager()
 
     job = job_manager.get_job(job_id)
--- a/tests/test_background_jobs.py
+++ b/tests/test_background_jobs.py
@@ -144,6 +144,23 @@ def test_list_jobs_tool_rejects_non_string_status():
         assert "Invalid status type" in result["error"]
 
 
+def test_cancel_job_tool_rejects_non_boolean_delete():
+    """delete must be a real boolean, not a truthy string or integer."""
+    from sktime_mcp.tools.job_tools import cancel_job_tool
+
+    job_manager = get_job_manager()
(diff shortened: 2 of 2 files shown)


Shashankss1205:
Thanks for the careful repro. The MCP Python SDK we depend on (now 1.28.x) validates tool arguments against the declared inputSchema before dispatch, and cancel_job.delete is declared boolean: on current main a call with delete='false' is rejected with 'Input validation error' before cancel_job_tool runs. The guard would only affect direct Python calls, which are not a supported surface, so closing this along with #359.

PR #363, opened 2026-04-29
FIX: validate formatting control flag types

### LINKED ISSUE #362: [BUG] format_time_series_tool accepts non-boolean control flags and mutates data unexpectedly
## Summary
`format_time_series_tool` currently accepts non-boolean values for `auto_infer_freq`, `fill_missing`, and `remove_duplicates`. Because the tool relies on Python truthiness, malformed values like strings can silently enable mutations that the caller did not request.

## Reproduction
Examples reproduced on current `main`:

```python
format_time_series_tool(..., remove_duplicates="false", fill_missing=False, auto_infer_freq=False)
# success=True
# duplicates are still removed
```

```python
format_time_series_tool(..., remove_duplicates=False, fill_missing="false", auto_infer_freq=False)
# success=True
# missing values are still filled
```

```python
format_time_series_tool(..., auto_infer_freq="false")
# truthy string sends the tool down frequency inference paths
```

## Why it matters
This is an MCP-facing reliability issue. Tool callers and LLM agents can pass stringified booleans, and the current behavior may silently rewrite user data instead of returning a validation error.

## Suggested fix
Validate that `auto_infer_freq`, `fill_missing`, and `remove_duplicates` are real booleans and return structured `success=False` errors for invalid types.

A small sibling fix may also make sense for `auto_format_on_load_tool(enabled=...)`, which currently accepts values like `"false"` and reports confusing success messages.

biru-codeastromer:
## Summary
- reject non-boolean `auto_infer_freq`, `fill_missing`, and `remove_duplicates` values in `format_time_series_tool`
- reject non-boolean `enabled` values in `auto_format_on_load_tool`
- add focused regression tests for each validation path

## Why
Formatting tools currently rely on Python truthiness. Values like `"false"` can silently remove duplicates, fill missing values, or produce confusing success messages instead of returning a structured validation error.

## Testing
- `.venv/bin/ruff check .`
- `.venv/bin/pytest tests/test_format_tools.py -q`
- `.venv/bin/pytest -q`

Closes #362

### DIFF
--- a/src/sktime_mcp/tools/format_tools.py
+++ b/src/sktime_mcp/tools/format_tools.py
@@ -9,6 +9,17 @@ from typing import Any
 from sktime_mcp.runtime.executor import get_executor
 
 
+def _validate_boolean_flag(flag_name: str, value: Any) -> dict[str, Any] | None:
+    """Return a structured error when a formatting control flag is not boolean."""
+    if isinstance(value, bool):
+        return None
+
+    return {
+        "success": False,
+        "error": (f"Invalid {flag_name} type '{type(value).__name__}'. Expected a boolean value."),
+    }
+
+
 def format_time_series_tool(
     data_handle: str,
     auto_infer_freq: bool = True,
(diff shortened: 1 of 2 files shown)


Shashankss1205:
Since April the formatting surface was consolidated: format_time_series_tool is no longer a registered MCP tool (formatting is transform_data(action='format'), which calls the executor directly), and the MCP SDK now validates arguments against the boolean schema before dispatch, so string booleans are rejected on current main. auto_format_on_load is deprecated and will be removed rather than hardened. Closing as obsolete along with #362.

PR #365, opened 2026-04-29
FIX: validate list_available_data is_demo input

### LINKED ISSUE #364: [BUG] list_available_data_tool accepts invalid is_demo values and returns misleading empty results
## Summary
`list_available_data_tool` currently accepts non-boolean values for `is_demo` and returns `success=True` with empty results instead of a validation error.

## Reproduction
Examples reproduced on current `main`:

```python
list_available_data_tool("false")
# {'success': True, 'system_demos': [], 'active_handles': [], 'total': 0}
```

```python
list_available_data_tool(1)
# {'success': True, 'system_demos': [], 'active_handles': [], 'total': 0}
```

```python
list_available_data_tool([])
# {'success': True, 'system_demos': [], 'active_handles': [], 'total': 0}
```

## Why it matters
This is an MCP-facing reliability issue. Tool callers and LLM agents can pass malformed boolean-like values, and the current behavior implies that no data exists instead of surfacing invalid input.

## Suggested fix
Require `is_demo` to be `True`, `False`, or `None`, and return a structured `success=False` error for any other type.

biru-codeastromer:
## Summary
- reject non-boolean `is_demo` values in `list_available_data_tool`
- return structured validation errors instead of misleading empty success responses
- add focused regression tests for invalid and valid `is_demo` paths

## Why
`list_available_data_tool` currently accepts malformed values like `"false"`, `1`, or `[]` and returns `success=True` with empty results. That behavior is misleading for MCP clients and LLM agents because it implies that no data exists instead of surfacing invalid input.

## Testing
- `.venv/bin/ruff check .`
- `.venv/bin/pytest tests/test_list_available_data.py -q`
- `.venv/bin/pytest -q`

Closes #364

### DIFF
--- a/src/sktime_mcp/tools/list_available_data.py
+++ b/src/sktime_mcp/tools/list_available_data.py
@@ -25,6 +25,15 @@ def list_available_data_tool(is_demo: bool | None = None) -> dict[str, Any]:
     """
     executor = get_executor()
 
+    if is_demo is not None and not isinstance(is_demo, bool):
+        return {
+            "success": False,
+            "error": (
+                f"Invalid is_demo type '{type(is_demo).__name__}'. "
+                "Expected a boolean value or None."
+            ),
+        }
+
     system_demos = []
     active_handles = []
 
--- a/tests/test_list_available_data.py
+++ b/tests/test_list_available_data.py
@@ -1,0 +1,47 @@
+"""Tests for list_available_data tool validation."""
+
+import sys
+
+sys.path.insert(0, "src")
+
+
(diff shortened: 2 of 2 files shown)


Shashankss1205:
Since this was opened the server moved to SDK-level input validation: every call is checked against the tool's inputSchema before dispatch, and list_available_data declares is_demo as boolean, so 'false', 1 or [] now return a structured 'Input validation error' instead of an empty success. That makes the in-function guard redundant for MCP clients, so closing this and #364; we would rather keep type checking in one place.

PR #366, opened 2026-04-29
[FIX] : Add substring fallback for tag suggestions in list_estimators

Rohitstwt:
#### Reference Issues/PRs
Related to #35 (early adopter feedback - tag suggestion returning None for partial key matches)

#### What does this implement/fix?

When an invalid tag key like `pred_int` is passed to `list_estimators_tool`, the existing code returns `suggestions: None` even though `capability:pred_int` exists in the registry.

**Root cause:** `difflib.get_close_matches` scores `pred_int` vs `capability:pred_int` below the 0.6 cutoff because the `capability:` prefix makes them too dissimilar.

**Before:**
```python
list_estimators_tool(tags={"pred_int": True})
# {'success': False, 'suggestions': {'pred_int': None}}
```

**After:**
```python
list_estimators_tool(tags={"pred_int": True})
# {'success': False, 'suggestions': {'pred_int': ['capability:pred_int', 'capability:pred_int:insample']}}
```

**Fix:** adds a `_suggest_tag` helper that first tries difflib, then falls back to substring matching when difflib finds nothing.

Also fixes the docstring which listed only 6 of 9 valid task types (missing `parameter_estimation`, `splitting`, `network`).

No. Uses only `difflib` from the Python standard library, which was already imported.

- The `_suggest_tag` helper in `src/sktime_mcp/tools/list_estimators.py` and whether the substring fallback logic is sound
- The 7 new tests in `tests/test_list_estimators.py` covering invalid task errors, invalid tag errors, substring suggestions, and valid inputs passing cleanly

#### Any other comments?
Discovered while doing early adopter testing as part of #35.

##### For all contributions
- [ ] Optionally, I've updated sktime's CODEOWNERS to receive notifications about future changes to these files.

### DIFF
--- a/src/sktime_mcp/tools/list_estimators.py
+++ b/src/sktime_mcp/tools/list_estimators.py
@@ -9,6 +9,15 @@ from typing import Any
 from sktime_mcp.registry.interface import get_registry
 
 
+def _suggest_tag(key: str, valid_tag_keys: set) -> list:
+    """Suggest valid tag keys for an invalid key using difflib and substring fallback."""
+    close = difflib.get_close_matches(key, valid_tag_keys, n=2, cutoff=0.6)
+    if close:
+        return close
+    # fallback: substring match e.g. "pred_int" -> "capability:pred_int"
+    return [t for t in valid_tag_keys if key in t]
+
+
 def list_estimators_tool(
     task: str | None = None,
     tags: dict[str, Any] | None = None,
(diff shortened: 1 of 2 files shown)


Shashankss1205:
The underlying bug still exists, it just moved: list_estimators was removed in #463 and the same difflib-only suggestion code now lives in src/sktime_mcp/tools/query_registry.py (around lines 104-113), where tags={'pred_int': true} still yields suggestions {'pred_int': null}. Could you re-target this PR to query_registry.py (your _suggest_tag helper can be reused as is), drop the list_estimators docstring change, and add a test? Then it should be a quick merge.

Me-Priyank:
Ran into this one, the file moved to `query_registry.py` in `d19e543` (#463), so the patch doesn't apply any more. The bug is still there though: `query_registry.py:104-115` has the identical block.

Confirmed the cutoff numerically, `SequenceMatcher(None, "pred_int", "capability:pred_int").ratio()` is 0.5926, just under `difflib`'s 0.6, so you get back `suggestions: {"pred_int": null}` with nothing to retry on.

One thought if you rebase: `valid_tag_keys` is a set, so the substring matches come back unordered and uncapped, `sorted(...)[:3]` might be safer for a short key like `"y"`.

PR #374, opened 2026-04-29
[FIX] validate input types in describe_estimator_tool and search_estimators_tool

### LINKED ISSUE #373: describe_estimator_tool and search_estimators_tool do not validate input types — non-string inputs cause unhandled exceptions[BUG]
describe_estimator_tool and search_estimators_tool do not validate the runtime type of their inputs before use. passing a non-string value causes an unhandled python exception instead of a structured error response. this is inconsistent with other tools in the codebase that return {"success": False, "error": "..."}

**to reproduce**

from sktime_mcp.tools.describe_estimator import describe_estimator_tool, search_estimators_tool

print(describe_estimator_tool(123))
# AttributeError: 'int' object has no attribute 'lower'

print(describe_estimator_tool(["ARIMA"]))
# TypeError: unhashable type: 'list'

print(search_estimators_tool(query=123))
# AttributeError: 'int' object has no attribute 'lower'

print(search_estimators_tool(query=""))
# returns {"success": True, "results": [], "count": 0} — should be a validation error

**expected behavior**

all four calls should return {"success": False, "error": "..."} instead of raising exceptions or returning misleading empty results.

**proposed fix**

add type and value checks at the top of both functions before any registry call or string method is called.

Agrim-Dubey:
Fixes #373

added input type validation in `describe_estimator_tool` and `search_estimators_tool`. non-string or empty string inputs now return `{"success": False, "error": "..."}` instead of raising unhandled exceptions like `AttributeError` or `TypeError`.

changes made:
- `describe_estimator_tool`: validate `estimator` is a non-empty string before any registry call
- `search_estimators_tool`: validate `query` is a non-empty string before any registry call

### DIFF
--- a/src/sktime_mcp/tools/describe_estimator.py
+++ b/src/sktime_mcp/tools/describe_estimator.py
@@ -39,6 +39,15 @@ def describe_estimator_tool(estimator: str) -> dict[str, Any]:
             ...
         }
     """
+    if not isinstance(estimator, str) or not estimator.strip():
+        return {
+            "success": False,
+            "error": (
+                f"'estimator' must be a non-empty string, "
+                f"got {type(estimator).__name__!r}. example: \"ARIMA\""
+            ),
+        }
+
     registry = get_registry()
     tag_resolver = get_tag_resolver()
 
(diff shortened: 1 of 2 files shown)


Shashankss1205:
Both functions it patches (describe_estimator_tool, search_estimators_tool) were removed in #463 in favour of describe_component/query_registry, and the MCP layer now validates arguments against each tool's inputSchema, so a non-string name is rejected with a structured error before the tool runs. Closing as obsolete.

PR #385, opened 2026-04-29
FIX: filter unsupported demo datasets from discovery

### LINKED ISSUE #384: [BUG] demo dataset discovery advertises non-forecasting datasets that high-level workflows cannot use
## Summary
The runtime currently auto-discovers all `sktime.datasets.load_*` functions and advertises them through demo dataset discovery.

After rejecting tuple-based supervised datasets in `load_dataset`, discovery still lists non-forecasting datasets like `basic_motions`, `arrow_head`, and `gunpoint` as available demos.

## Reproduction
On current code, `list_available_data_tool(is_demo=True)` includes datasets such as:
- `basic_motions`
- `arrow_head`
- `gunpoint`

But those datasets are not actually usable in the current forecasting-oriented high-level workflow.

## Why it matters
This leaves the MCP surface misleading at the discovery layer. Agents are told these datasets are valid demo choices, only to hit a failure later when they try to load them.

## Suggested fix
Filter discovery so only demo datasets compatible with the current forecasting-oriented workflow are advertised.

biru-codeastromer:
## Summary
- filter auto-discovered demo datasets to forecasting-compatible loaders in `Executor.list_datasets()`
- stop advertising supervised `(X, y)` demo datasets like `basic_motions`, `arrow_head`, and `gunpoint`
- add focused regression tests for executor-level and MCP-facing demo discovery

## Why
The runtime currently advertises all `sktime.datasets.load_*` functions as demo datasets. That includes supervised/classification loaders that are not usable in the current forecasting-oriented high-level workflow, which makes MCP discovery misleading.

## Notes
This PR keeps its scope to **discovery filtering** only. Runtime load-time handling for unsupported datasets is separate.

## Testing
- `.venv/bin/ruff check .`
- `.venv/bin/pytest tests/test_demo_dataset_discovery.py -q`
- `.venv/bin/pytest -q`

Closes #384

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -40,6 +40,38 @@ def _discover_demo_datasets() -> dict:
 
 
 DEMO_DATASETS = _discover_demo_datasets()
+_SUPPORTED_DEMO_DATASETS: list[str] | None = None
+
+
+def _is_supported_demo_dataset(name: str, module_path: str) -> bool:
+    """Return whether a discovered demo dataset matches current forecasting-oriented support."""
+    try:
+        parts = module_path.rsplit(".", 1)
+        module = __import__(parts[0], fromlist=[parts[1]])
+        loader = getattr(module, parts[1])
+        signature = inspect.signature(loader)
+    except Exception:
+        logger.warning(
+            "Skipping demo dataset '%s' because its loader could not be inspected.", name
+        )
+        return False
+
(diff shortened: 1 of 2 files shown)


Shashankss1205:
Hey @biru-codeastromer thanks but I think this is not needed, We can add support steadily for the other estimators rather than removing the datasets. Good catch though :)

biru-codeastromer:
> Hey @biru-codeastromer thanks but I think this is not needed, We can add support steadily for the other estimators rather than removing the datasets. Good catch though :)

Understood , no worries.

PR #389, opened 2026-04-29
FIX: reject unsupported demo datasets in export_code

### LINKED ISSUE #388: [BUG] export_code accepts unsupported demo datasets for fit examples and generates misleading forecasting code
## Summary
`export_code_tool(..., include_fit_example=True, dataset=...)` currently accepts demo datasets that are not compatible with the forecasting-oriented fit example template.

## Reproduction
On current `main`, this succeeds:

```python
export_code_tool(handle, include_fit_example=True, dataset="basic_motions")
```

and generates code like:

```python
from sktime.datasets import load_basic_motions
y = load_basic_motions()
model.fit(y)
```

## Why it matters
`basic_motions` is not a forecasting demo dataset in the current high-level workflow. Generated code is trusted output, so producing misleading example code is worse than rejecting the request.

## Suggested fix
When `include_fit_example=True`, validate that the requested dataset is compatible with the forecasting-oriented fit example template and return a clear error for unsupported demo datasets.

biru-codeastromer:
## Summary
- reject unsupported supervised demo datasets when `export_code` is asked to include a forecasting fit example
- keep valid forecasting datasets like `airline` working as before
- add focused regression tests for supported and unsupported dataset paths

## Why
`export_code_tool(..., include_fit_example=True, dataset=...)` currently trusts the raw discovered demo dataset map and can generate misleading forecasting example code for datasets like `basic_motions`.

## Testing
- `.venv/bin/ruff check .`
- `.venv/bin/pytest tests/test_codegen.py -q`
- `.venv/bin/pytest -q`

Closes #388

### DIFF
--- a/src/sktime_mcp/tools/codegen.py
+++ b/src/sktime_mcp/tools/codegen.py
@@ -4,6 +4,7 @@ Code generation tool for sktime MCP.
 Generates Python code to recreate estimators and pipelines.
 """
 
+import inspect
 import keyword
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


Shashankss1205:
not needed

Issue #400, opened 2026-04-30
[ENH] Add diagnose_residuals_tool to give agents statistical reasoning for model failures


DEEP-600:
## The Problem What and why

Going through the sktime-mcp tool surface, I noticed the agentic 
loop has no feedback mechanism after a poor evaluation score. The 
agent gets MAPE = 25% and has nothing to work with beyond trying 
another model.

Human forecasters look at residuals at this point — ACF for missed 
seasonality, normality checks, bias direction. Agents can't look at 
plots, so this tool runs those same checks and returns structured 
text they can reason over.

## The Proposal

I propose adding a new MCP tool: `diagnose_residuals_tool`.

This tool computes the in-sample residuals of a fitted model and runs 
statistical tests to explain the model's errors to the LLM in a 
structured, actionable format.

**How it fits the current architecture:**

1. Follows the exact pattern of `evaluate.py`: accepts `estimator_handle` 
   and `dataset`
2. Reloads data via `executor.load_dataset()`
3. Retrieves the fitted model via `executor._handle_manager.get_instance()`
4. Calls `instance.predict_residuals(y)` (defined on `BaseForecaster`) to 
   compute errors
5. Uses `statsmodels` and `scipy` (already in `pyproject.toml`) for 
   Ljung-Box, Shapiro-Wilk, and bias tests

**Example output returned to the LLM:**

```json
{
  "success": true,
  "diagnostics": {
    "bias": {
      "mean_error": -4.2,
      "status": "consistently under-forecasting"
    },
    "autocorrelation": {
      "ljung_box_passed": false,
      "significant_lags": [12, 24]
    },
    "normality": {
      "shapiro_passed": false,
      "p_value": 0.01
    }
  },
  "llm_hint": "Residuals show significant autocorrelation at lag 12. 
  The model failed to capture annual seasonality. Consider switching 
  to SARIMA or adding a Deseasonalizer pipeline."
}
```

The `llm_hint` field is key — it translates raw statistics directly 
into actionable reasoning for the agent's next ReAct step.

This tool completes the feedback loop discussed in #386 and complements 
the evaluation work in #369.

##  Questions...
1. **Residual calculation:** `BaseForecaster.predict_residuals(y)` covers 
   the broad case cleanly. Should we add a manual fallback 
   (`y - instance.predict(fh=y.index)`) for edge cases with complex 
   pipelines or neural estimators?

[...]

Issue #386, opened 2026-04-29
[ENH] Agentic forecaster workflow: missing tools for iterative candidate evaluation


kpal002:
## Context

I am working on an agentic forecaster prototype for the ESoC 2026 sktime track (see sktime/sktime#9721). The idea: an LLM agent runs a ReAct loop, calling sktime-mcp tools to explore forecaster candidates, score them on a holdout
window, and commit the best one. While doing this I ran into several gaps in the current tool surface.

## What I tried

I implemented an in-process version first (prototype repo: https://github.com/kpal002/sktime-agentic-forecaster) to establish the tool contract, then tried to replicate it against the sktime-mcp server. Three things are missing:

### 1. No data fingerprinting tool

The agent's first action should be to inspect the series — length, frequency, trend slope, missingness, candidate seasonal period — so it can make an informed choice before calling `list_estimators`. The existing tools expose no
equivalent. `load_data_source` + `format_time_series` prepare the data but return no summary statistics for LLM reasoning.

**Suggested tool:** `describe_data(handle)` → returns a dict with at least:
```json
{
  "length": 144,
  "frequency": "ME",
  "n_missing": 0,
  "mean": 280.3,
  "std": 78.4,
  "trend_slope_per_step": 1.2,
  "candidate_seasonal_period": 12
}
```
2. No single-candidate holdout scoring
evaluate_estimator_tool (and the proposed #369 compare_estimators_tool) run full cross-validation across many estimators at once. That is the right design for a one-shot batch call, but it does not support the iterative agent pattern where the LLM:

- Fits one candidate with specific params
- Observes the score
- Decides whether to try another variant
- Eventually commits

What is missing is a lighter-weight tool: fit a single named estimator on the in-sample portion of a data handle, score it against a held-out tail, and return the scalar metric — without running full CV or requiring the LLM to
batch all candidates upfront.

Suggested tool: score_on_holdout(handle, estimator_name, params, holdout, metric)

3. No "commit and finalize" tool
Once the agent has chosen a winner, it needs to: refit on the full training series, store the fitted model as a named handle, and surface a rationale. The current fit_predict tool trains and predicts in one call — it does not
support the "fit now, predict later at a different horizon" separation that a committed agent model needs.

A commit_estimator(handle, estimator_name, params, rationale) tool that refits on the full series and returns a persistent model handle would close this gap.

Why this matters
[...]

PR #402, opened 2026-04-30
feat: add diagnose_residuals_tool for agentic residual diagnostics

DEEP-600:
Implements #400.

## What and why

Reading through the existing tool surface and #386, I noticed the 
agentic loop has no feedback mechanism after a poor evaluation score. 
The agent gets MAPE = 25% and has nothing to work with beyond trying 
another model.

Human forecasters look at residuals at this point — ACF for missed 
seasonality, normality checks, bias direction. Agents can't look at 
plots, so this tool runs those same checks and returns structured text 
they can reason over.
## What's in this PR

- `src/sktime_mcp/tools/diagnose.py` — the new tool
- `__init__.py` and `server.py` updated to expose it
- `tests/test_diagnose.py` — three test cases

## How it works

Takes an `estimator_handle` and `dataset`, reloads the data the same 
way `evaluate.py` does, pulls the fitted instance from 
`_handle_manager`, and runs three tests on the residuals:

- Ljung-Box (statsmodels) — catches missed seasonality
- Shapiro-Wilk (scipy) — catches non-normality  
- Mean bias — catches systematic under/over-forecasting

No new dependencies — `statsmodels` and `scipy` are already in 
`pyproject.toml`.

## Example output

```json
{
  "success": true,
  "diagnostics": {
    "bias": {"mean_error": -4.2, "status": "consistently over-forecasting"},
    "autocorrelation": {"ljung_box_passed": false, "significant_lags": [12, 24]},
    "normality": {"shapiro_passed": false, "p_value": 0.01}
  },
  "llm_hint": "Residuals show significant autocorrelation at lags [12, 24]. 
  This may indicate missed annual seasonality. Consider switching to SARIMA 
  or adding a Deseasonalizer pipeline."
}
```

## A couple of things I'd like feedback on

- I used `predict_residuals(y)` as primary with a manual fallback for 
  complex pipelines — is that the right call or overkill?
- Kept heteroskedasticity out of scope for now, happy to add in a 
  follow-up if that's useful.

### DIFF
--- a/src/sktime_mcp/server.py
+++ b/src/sktime_mcp/server.py
@@ -37,6 +37,7 @@ from sktime_mcp.tools.data_tools import (
     release_data_handle_tool,
 )
 from sktime_mcp.tools.describe_estimator import describe_estimator_tool
+from sktime_mcp.tools.diagnose import diagnose_residuals_tool
 from sktime_mcp.tools.evaluate import evaluate_estimator_tool
 from sktime_mcp.tools.fit_predict import (
     fit_predict_async_tool,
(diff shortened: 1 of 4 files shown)


Shashankss1205:
This PR depends on evaluate_estimator_tool, fit_predict_tool and the old server.py dispatcher, all replaced in the tool consolidation (#460), and #270 settled on keeping the surface at 26 general tools rather than adding task-specific ones. Residual diagnostics can be run today with call_method(handle, 'predict_residuals') plus run_command for the statistical tests. Closing.

PR #413, opened 2026-04-30
[BUG] fix auto_format_on_load crash for RangeIndex time series (

### LINKED ISSUE #390: [BUG] auto_format_on_load crashes internally for valid RangeIndex time series
## Summary
When `load_data_source` receives valid integer-indexed pandas data with auto-format enabled, the request succeeds but the auto-format step logs an internal failure:

```
Auto-formatting failed: 'RangeIndex' object has no attribute 'freq'\n```\n\n## Reproduction\n```python\nfrom sktime_mcp.runtime.executor import Executor\n\nExecutor().load_data_source({\n    "type": "pandas",\n    "data": {"y": [1, 2, 3, 4]},\n})\n```\n\n## Current behavior\n- returns `success=True`\n- auto-formatting falls back after logging an internal failure\n- valid integer-indexed time series do not go through the normal formatting path\n\n## Expected behavior\n- integer / `RangeIndex` inputs should be treated as valid forecasting data\n- auto-formatting should skip datetime-only frequency inference for non-datetime indexes\n- metadata should preserve a sensible frequency label such as `Integer` instead of degrading or failing\n\n## Why this matters\nThe adapters and validation layer already allow integer-style indexes. The formatter should not assume datetime-only frequency metadata, otherwise a valid workflow partially fails behind the scenes and becomes harder for MCP clients to trust.

eyasir2047:
#### Reference Issues/PRs
Fixes #390

I ran into this while testing  load_data_source with a simple dict-based dataset (no datetime column). The auto-format step was crashing internally with 'RangeIndex' object has no attribute 'freq' — even though the request returned success=True, the formatting silently failed behind the scenes.

The root cause is in  format_data_handle(): it calls y.index.freq, pd.infer_freq() and pd.date_range() without checking whether the index is actually datetime-like. A plain integer/RangeIndex doesn't have a .freq attribute, so it blows up.

The fix is a simple guard at the start of the frequency inference block — if the index is a  RangeIndex, we skip all the datetime logic and just label the frequency as "Integer". Nothing else changes for datetime-indexed data.

No, just uses pd.RangeIndex which is already available.

Mostly the guard condition around line 790 in executor.py— want to make sure it covers all the non-datetime index cases cleanly.

#### Any other comments?
Added 3 tests covering the unit-level fix, metadata output, and the full end-to-end load_data_sourcecall. Full test suite passes, ruff and format checks are clean.

##### For all contributions
[x] I've added unit tests and made sure they pass locally.

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -788,46 +788,54 @@ class Executor:
 
         # 3. Infer and set frequency
         if auto_infer_freq:
-            freq = y.index.freq
-
-            if freq is None:
-                # Try to infer
-                freq = pd.infer_freq(y.index)
+            # Guard: integer / RangeIndex data should not go through
+            # datetime-specific frequency inference (fixes #390).
+            if isinstance(y.index, pd.RangeIndex) or not hasattr(y.index, "freq"):
+                changes_made["frequency_set"] = True
+                changes_made["frequency"] = "Integer"
+            else:
+                freq = y.index.freq
 
                 if freq is None:
-                    # Manual inference
-                    time_diffs = y.index.to_series().diff().dropna()
-                    if len(time_diffs) > 0:
-                        most_common_diff = time_diffs.mode()[0]
+                    # Try to infer
+                    freq = pd.infer_freq(y.index)
 
-                        if most_common_diff == pd.Timedelta(days=1):
-                            freq = "D"
-                        elif most_common_diff == pd.Timedelta(hours=1):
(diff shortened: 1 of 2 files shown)


biru-codeastromer:
@eyasir2047  You made a duplicate PR to a already linked PR. Please check the issue if there are any attached linked PR's

eyasir2047:
@biru-codeastromer  Thanks for letting me know! I'll take a look at the linked PR and close this one if it already covers the fix.

Shashankss1205:
Duplicated

PR #415, opened 2026-04-30
[BUG] Fix server stability: JSON serialization, KeyboardInterrupt, & tool input validations

### LINKED ISSUE #249: [BUG] cleanup_old_jobs_tool accepts zero/negative max_age_hours silently deletes all jobs

### LINKED ISSUE #335: [BUG] Hardcoded 500-character docstring truncation limits LLM reasoning
### Description
In `src/sktime_mcp/registry/interface.py`, the `docstring` for `EstimatorNode` is hardcoded to truncate at 500 characters. 
```python
"docstring": (self.docstring[:500] if self.docstring else None)
```

### Why it's an issue
Large Language Models (LLMs) rely on these docstrings to understand how to configure estimators. Many sktime models have detailed docstrings where the most critical information (parameter descriptions) appears after the first 500 characters. Truncating this early prevents agents from correctly "reasoning" about the model's capabilities and hyperparameters.

### Suggested Fix
Increase the truncation limit to 2000+ characters or implement a logic that prioritizes the "Parameters" section of the docstring.

### LINKED ISSUE #373: describe_estimator_tool and search_estimators_tool do not validate input types — non-string inputs cause unhandled exceptions[BUG]

### LINKED ISSUE #390: [BUG] auto_format_on_load crashes internally for valid RangeIndex time series

### LINKED ISSUE #394: [BUG]Server crashes with traceback on Ctrl+C
Hi team,

I was testing the local setup on Windows 11. When running the server via CLI (`python -m sktime_mcp.server`), pressing `Ctrl+C` to terminate the process does not shut down gracefully. 

Instead of a clean exit, it throws an unhandled `KeyboardInterrupt` which cascades into an `anyio.WouldBlock` and `asyncio.exceptions.CancelledError` traceback.

**Steps to reproduce:**
1. `pip install -e ".[dev]"`
2. Run `python -m sktime_mcp.server`
3. Press `Ctrl+C`

**Traceback:**
<details>
<summary>Click to expand</summary>
```python
Traceback (most recent call last):
  File "E:\ESOC\sktime-mcp\venv\Lib\site-packages\anyio\streams\memory.py", line 117, in receive
    return self.receive_nowait()
    # ... [cascading anyio/asyncio errors] ...
  File "C:\Users\abdul\AppData\Local\Programs\Python\Python312\Lib\asyncio\runners.py", line 123, in run
    raise KeyboardInterrupt()
KeyboardInterrupt

### LINKED ISSUE #400: [ENH] Add diagnose_residuals_tool to give agents statistical reasoning for model failures

### LINKED ISSUE #401: [BUG] numpy.float64 values in prediction results cause JSON serialization failure
**Describe the bug**
In `src/sktime_mcp/runtime/executor.py` line 171, predictions are 
returned via `.to_dict()` without converting numpy types to Python 
native types. This causes JSON serialization to fail in MCP responses 
since `numpy.float64` is not JSON serializable.

**Location**
`src/sktime_mcp/runtime/executor.py:171`

**Minimal reproduction**
Run any fit_predict call and inspect the returned dict — values will 
be numpy.float64 instead of Python float.

**Expected**
Prediction values should be Python native float.

**Actual**
Prediction values are numpy.float64, which fails JSON serialization.

**Proposed fix**
Convert values at dict construction:
result = {str(k): float(v) if hasattr(v, 'item') else v 
          for k, v in predictions_copy.to_dict().items()}

### LINKED ISSUE #405: [BUG] Classification workflows are discoverable but not executable end-to-end via MCP
Hey team, sorry for not following the standard bug report template, this felt a little different from a bug report and more towards developer experience so I'm making a blank issue instead.
(For transparency: I used an LLM to rephrase and better structure my issue draft)

So while testing sktime-mcp with Claude, I found that forecasting works end-to-end, but two other core sktime workflows (classification and forecast evaluation) are unreachable through MCP tools alone.

The MCP surface implies support that doesn't exist in practice, which causes unreliable agentic behavior and user confusion.

**What works**

- `list_estimators(task="classification")` returns 66 valid classifiers
- `describe_estimator` works as expected
- `instantiate_estimator` works
- Forecasting workflows work end-to-end with built-in datasets

**Gap 1: Classification workflows are discoverable but not executable**

**No usable built-in classification datasets**
`list_datasets` only returns 6 forecasting datasets. Attempting common sktime datasets like `ArrowHead` fails with `"Unknown dataset"`.

**No reliable path to load classification data**
URL loading is fragile and format-sensitive; standard sktime dataset URLs return 404. There's no documented schema for panel/classification data input, and the `pandas` adapter requires a DataFrame (not a list or dict) which is undocumented.

**Data validation blocks valid panel data**
Multi-instance panel data fails with `"Duplicate time indices found"` because the validator treats panel data as a single time series. Error messages don't explain the expected format for classification tasks.

**`fit_predict` is forecasting-only**
No equivalent execution tool exists for classification, clustering, or regression tasks.

When constrained to MCP tools only, the agent correctly discovers and instantiates a classifier, then attempts 8+ fallback strategies (URL variants, synthetic pandas data, dict input, list input) and still cannot complete the workflow.

**Gap 2: No forecast evaluation tools**

After completing a forecasting workflow (Airline dataset, Detrender -> Deseasonalizer -> ARIMA, 36-month horizon), there's no way to evaluate it. There are no MCP tools for train/test splitting, backtesting, cross-validation, or metric computation (MAE, RMSE, MAPE, etc.).

The agent correctly diagnosed the gap and fell back to writing raw Python, which defeats the purpose of an agentic integration. A user can produce a forecast but has no way to measure whether it's any good, entirely within MCP.

Session transcript: https://claude.ai/share/0c8119c7-07f8-4457-9b7e-2c597dace295

So we need to either complete the surface (classification datasets, panel data loader, a `fit_classify` equivalent, evaluation tools) or narrow it honestly and don't expose tasks that have no execution path yet.

**To reproduce**

Connect sktime-mcp to any Claude session and run:
1. "Train a time series classifier on a built-in dataset using only MCP tools", agent discovers and instantiates a classifier but hits a dead end at data loading
2. "Evaluate the forecast you just made" - agent confirms no evaluation tools exist and falls back to Python

Happy to help test fixes or contribute to either gap.

### LINKED ISSUE #407: Add JSON Schema constraints for task and cv_folds in MCP server
### Is your feature request related to a problem? Please describe.

Yes. The tool input schemas in `server.py` lack constraints, which causes LLMs to send invalid values.

- Invalid `task` values (e.g., "forecasting_model" instead of "forecasting")
- Invalid `cv_folds` values (negative or very large numbers)

---

### Describe the solution you'd like

Add JSON Schema constraints:

For `task`:
```python
"task": {
    "type": "string",
    "enum": ["forecasting", "classification", "regression", "transformation", "clustering"]
}

### LINKED ISSUE #408: [ENH] Add run_tools_batch MCP tool for read-only agent workflows
## Motivation
Agentic workflows usually require multiple read-only calls in sequence (e.g., list estimators -> describe candidates -> list available tags/data).
This creates avoidable round-trips and latency for MCP clients.

## Proposal
Add a new MCP tool: `run_tools_batch`

### Scope (MVP)
- Accept a list of operations: `{tool: str, arguments: dict}`
- Execute only read-only tools in batch:
  - `list_estimators`
  - `describe_estimator`
  - `get_available_tags`
  - `list_available_data`
- Return per-operation structured results with index + tool + success/error
- Reject unsupported tools with clear validation errors

### Out of scope for MVP
- Mutating tools (`instantiate_*`, `fit_predict*`, `release_*`, jobs mutation)
- Parallel execution or DAG semantics

## Acceptance criteria
- New tool appears in `list_tools` with schema
- `call_tool` routes `run_tools_batch` requests
- Unit tests cover success path, unknown tool in batch, malformed operation entries
- Output is JSON-serializable and stable for agents

If maintainers agree, I can open a focused PR with tests first.

### LINKED ISSUE #410: [BUG]Remove all references of mkdocs and shift everything to sphinx of readthedocs.
We earlier added docs in mkdocs format, but to remain consistent with the sktime specs, we shifted it mostly to readthedocs and sphinx. There are still some older references that have been left out during the refactor. It's very helpful to remove all such occurrences.

### LINKED ISSUE #416: [ENH] expose explicit data shape metadata for agent reasoning
## Summary
`load_data_source` and `load_data_source_async` currently return basic metadata such as rows, columns, dates, and dtypes, but they do not explicitly tell an agent whether a loaded handle is univariate or multivariate, whether exogenous variables are present, or what kind of index backs the series.

## Why this matters
Agentic workflows often need to decide which estimator family or tool path to take before fitting. Right now an agent has to infer too much from raw column names and dtypes.

Concrete gaps:
- univariate vs multivariate target is not explicit
- exogenous presence is not explicit
- exogenous width is not explicit
- index type is not explicit (`datetime` vs `period` vs `range` / `integer`)

This makes tool selection and model routing harder than it needs to be.

## Proposal
Augment data-handle metadata with explicit agent-friendly fields, for example:
- `target_scitype`
- `target_variates`
- `has_exog`
- `exog_variates`
- `index_type`
- `n_target_columns`
- `n_exog_columns`

## Expected benefit
This would make loaded data self-describing for LLM / MCP agents and reduce avoidable reasoning errors when selecting estimators or deciding whether exogenous-aware workflows are available.

SHIVANSH-ux-ys:
#### Reference Issues/PRs

Fixes #401
Fixes #394
Fixes #410
Fixes #373
Fixes #335
Fixes #249
Fixes #407
Fixes #408
Fixes #416 (Note: Overlaps with PR #417 by @biru-codeastromer, happy to defer to their implementation)
Fixes #405
Fixes #400
Fixes #390 (Note: Overlaps with PR #391 by @biru-codeastromer, happy to defer to their implementation)

This PR implements a series of critical stability improvements and bug fixes for the `sktime-mcp` server to ensure smoother agentic workflows:

1. **JSON Serialization (#401):** Fixed an issue where `pandas.Series.to_dict()` leaked non-serializable `numpy.float64` values into tool outputs. Implemented type casting to native Python `float` in `executor.py` so standard `json.dumps()` works without crashing.
2. **Graceful Shutdown (#394):** The server previously crashed with a messy traceback upon `Ctrl+C`. Wrapped the `main()` async loop in `server.py` with a `KeyboardInterrupt` catch for clean exits.
3. **Docs Migration (#410):** Removed outdated `mkdocs` build references from `dev-guide.md` and updated instructions to reflect the new Sphinx/ReadTheDocs pipeline.
5. **Tool Input Validation (#373 & #249):** Added robust input type checking to `describe_estimator_tool` and `search_estimators_tool` to prevent unhandled exceptions on non-string inputs. Also added positive-integer validation to `cleanup_old_jobs_tool` to prevent negative inputs from wiping all jobs.
4. **Docstring Context (#335):** Removed the hardcoded 500-character truncation limit in `describe_estimator` so that LLMs receive the full estimator context for better reasoning.

No new dependencies are introduced.

Reviewers can verify that the type-casting logic in `executor.py` properly covers all pandas-to-dict prediction edge cases.

#### Did you add any tests for the change?

No formal pytest cases were added, but local validation scripts were successfully run against `executor.py` and `json.dumps` to verify the serialization fix natively handles all numpy types.
6. **RangeIndex Formatting Crash (Fixes #390):** Fixed an internal crash where `auto_format_on_load` would fail with an AttributeError when formatting simple `RangeIndex` datasets without a datetime frequency.

### DIFF
--- a/docs/source/dev-guide.md
+++ b/docs/source/dev-guide.md
@@ -6,7 +6,7 @@ This guide explains how the project is structured, how to develop new features, 
 
 - Python 3.10+
 - pip
-- Optional: `mkdocs` if you want to build the documentation site
+- Optional: Sphinx if you want to build the documentation site
 
 ## Setup
 
@@ -114,8 +114,9 @@ Update `DEMO_DATASETS` in `src/sktime_mcp/runtime/executor.py` with a new loader
 If you want to build the docs site locally:
 
 ```bash
-pip install mkdocs
-mkdocs serve
+pip install -r docs/requirements.txt
+cd docs
+make html
 ```
 
-The config lives in `mkdocs.yml` and the docs content is under `docs/`.
+The config lives in `docs/source/conf.py` and the docs content is under `docs/source/`.
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -131,12 +131,16 @@ class Executor:
             return {"success": False, "error": f"Handle not found: {handle_id}"}
 
         try:
-            if fh is not None:
-                instance.fit(y, X=X, fh=fh)
-            elif X is not None:
-                instance.fit(y, X=X)
+            is_classifier = getattr(instance, "_estimator_type", None) == "classifier"
+            if is_classifier:
+                instance.fit(X, y)
             else:
-                instance.fit(y)
+                if fh is not None:
+                    instance.fit(y, X=X, fh=fh)
+                elif X is not None:
+                    instance.fit(y, X=X)
+                else:
+                    instance.fit(y)
 
             self._handle_manager.mark_fitted(handle_id)
             return {"success": True, "handle": handle_id, "fitted": True}
@@ -159,10 +163,14 @@ class Executor:
             return {"success": False, "error": "Estimator not fitted"}
 
         try:
-            if fh is None:
-                fh = list(range(1, 13))
+            is_classifier = getattr(instance, "_estimator_type", None) == "classifier"
+            if is_classifier:
+                predictions = instance.predict(X)
+            else:
+                if fh is None:
+                    fh = list(range(1, 13))
 
-            predictions = instance.predict(fh=fh, X=X) if X is not None else instance.predict(fh=fh)
+                predictions = instance.predict(fh=fh, X=X) if X is not None else instance.predict(fh=fh)
 
             if isinstance(predictions, pd.Series):
                 # Convert index to string to avoid JSON serialization issues with Period/DatetimeIndex
@@ -580,6 +588,8 @@ class Executor:
                 metadata["exog_columns"] = list(X.columns)
             # Inject column dtypes so LLMs can distinguish time index vs target
             metadata["dtypes"] = {col: str(dtype) for col, dtype in data.dtypes.items()}
+            # Expose explicit shape metadata for agent reasoning (#416)
+            metadata["shape"] = list(data.shape) if hasattr(data, "shape") else len(data)
             # Generate handle
             data_handle = f"data_{uuid.uuid4().hex[:8]}"
 
@@ -614,6 +624,7 @@ class Executor:
                     # Continue with unformatted data if formatting fails
             _final_meta = adapter.get_metadata().copy()
             _final_meta["dtypes"] = {col: str(dtype) for col, dtype in data.dtypes.items()}
+            _final_meta["shape"] = list(data.shape) if hasattr(data, "shape") else len(data)
             return {
                 "success": True,
                 "data_handle": data_handle,
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -802,11 +802,16 @@ class Executor:
 
         # 3. Infer and set frequency
         if auto_infer_freq:
-            freq = y.index.freq
+            freq = None
+            if not isinstance(y.index, (pd.DatetimeIndex, pd.PeriodIndex)):
+                changes_made["frequency_set"] = True
+                changes_made["frequency"] = "Integer"
+            else:
+                freq = getattr(y.index, "freq", None)
 
-            if freq is None:
-                # Try to infer
-                freq = pd.infer_freq(y.index)
+                if freq is None:
+                    # Try to infer
+                    freq = pd.infer_freq(y.index)
 
                 if freq is None:
                     # Manual inference
@@ -867,7 +872,7 @@ class Executor:
             "metadata": {
                 **data_info["metadata"],
                 "formatted": True,
-                "frequency": str(y.index.freq) if y.index.freq else changes_made.get("frequency"),
+                "frequency": str(y.index.freq) if getattr(y.index, "freq", None) else changes_made.get("frequency"),
                 "rows": len(y),
                 "start_date": str(y.index.min()),
                 "end_date": str(y.index.max()),
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -168,11 +168,14 @@ class Executor:
                 # Convert index to string to avoid JSON serialization issues with Period/DatetimeIndex
                 predictions_copy = predictions.copy()
                 predictions_copy.index = predictions_copy.index.astype(str)
-                result = predictions_copy.to_dict()
+                # Convert values to native Python types for JSON serialization
+                result = {k: float(v) if hasattr(v, "item") else v for k, v in predictions_copy.to_dict().items()}
             elif isinstance(predictions, pd.DataFrame):
                 predictions_copy = predictions.copy()
                 predictions_copy.index = predictions_copy.index.astype(str)
-                result = predictions_copy.to_dict(orient="list")
+                # Convert lists of values to native Python types
+                raw_dict = predictions_copy.to_dict(orient="list")
+                result = {k: [float(v_i) if hasattr(v_i, "item") else v_i for v_i in v_list] for k, v_list in raw_dict.items()}
             else:
                 result = predictions.tolist() if hasattr(predictions, "tolist") else predictions
 
--- a/src/sktime_mcp/server.py
+++ b/src/sktime_mcp/server.py
@@ -37,7 +37,7 @@ from sktime_mcp.tools.data_tools import (
     release_data_handle_tool,
 )
 from sktime_mcp.tools.describe_estimator import describe_estimator_tool
-from sktime_mcp.tools.evaluate import evaluate_estimator_tool
+from sktime_mcp.tools.evaluate import evaluate_estimator_tool, diagnose_residuals_tool
 from sktime_mcp.tools.fit_predict import (
     fit_predict_async_tool,
     fit_predict_tool,
@@ -376,6 +376,24 @@ async def list_tools() -> list[Tool]:
                     },
                 },
                 "required": ["estimator_handle", "dataset"],
+            },
+        ),
+        Tool(
+            name="diagnose_residuals_tool",
+            description="Diagnose model failures by calculating statistical metrics (MAE, RMSE, Mean Bias) on residuals.",
+            inputSchema={
+                "type": "object",
+                "properties": {
+                    "predictions": {
+                        "type": ["object", "array"],
+                        "description": "Forecasted values (dict or list)",
+                    },
+                    "actuals": {
+                        "type": ["object", "array"],
+                        "description": "Actual observed values (dict or list)",
+                    },
+                },
+                "required": ["predictions", "actuals"],
             },
         ),
         # -- Batch Execution -------------------------------------------------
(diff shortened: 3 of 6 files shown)


biru-codeastromer:
It looks like some of the issues you worked on already had linked PRs. It might help to check for existing contributions before starting, so effort isn’t duplicated.

PR #441, opened 2026-05-04
fix: raise error for missing explicit exogenous columns

Saloni-0465:
## Reference Issues/PRs
N/A

This PR fixes silent dropping of explicitly requested exogenous columns during data conversion.

Previously, if a config specified `exog_columns=["promo", "holiday"]` but only `promo` existed in the loaded data, `to_sktime_format` would silently use the available column and ignore the missing one. This could mislead users or LLM agents into thinking all requested covariates were used.

This PR updates `DataSourceAdapter.to_sktime_format` to validate explicit `exog_columns` before constructing `X`. If any requested exogenous column is missing, it raises a clear `ValueError` listing the missing columns and available columns. When all requested columns exist, their order is preserved exactly.

No.

- Whether missing explicit `exog_columns` should fail fast rather than warn
- Clarity of the error message returned to users and agents
- Whether the validation belongs in `DataSourceAdapter.to_sktime_format`

## Any other comments?
This is a small data-contract safety fix for custom data workflows. It helps prevent agents from making forecasting decisions under the false assumption that missing exogenous variables were used.

Validation:
- `python -m compileall src/sktime_mcp/data/base.py tests/test_data_exog_column_validation.py`
- Manual check: missing `holiday` in `exog_columns` raises a clear `ValueError`
- Manual check: `Executor.load_data_source` returns `success=False` with `error_type="ValueError"`
- Manual check: valid explicit exogenous columns are preserved in order

## PR checklist
For all contributions
- [ ] Optionally, I've updated CODEOWNERS.
- [x] I've added unit tests and made sure they pass locally where possible.

For new estimators
- [x] Not applicable

### DIFF
--- a/src/sktime_mcp/data/base.py
+++ b/src/sktime_mcp/data/base.py
@@ -79,8 +79,15 @@ class DataSourceAdapter(ABC):
 
             # Get exogenous variables if specified
             if exog_cols:
-                valid_exog_cols = [col for col in exog_cols if col in data.columns]
-                X = data[valid_exog_cols] if valid_exog_cols else None
+                missing_exog_cols = [col for col in exog_cols if col not in data.columns]
+                if missing_exog_cols:
+                    available_columns = ", ".join(repr(col) for col in data.columns)
+                    raise ValueError(
+                        f"Exogenous column(s) not found in data: {missing_exog_cols!r}. "
+                        f"Available columns: [{available_columns}]"
+                    )
+
+                X = data[exog_cols]
             else:
                 # Use all columns except target as exogenous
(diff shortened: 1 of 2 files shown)


Shashankss1205:
Verified on current main that a missing entry in exog_columns is silently dropped with success true; with this branch load_data_source returns a clear ValueError naming the missing and available columns. Full suite green. Merging. Follow-up: a bare string exog_columns is still iterated per character and should be coerced to a one-element list.

Shashankss1205:
Landed on main as 3afd1cd (squash-merged locally because the org policy rejects the merge API for this token). Thanks!

PR #443, opened 2026-05-04
fix: propagate default target column warnings

Saloni-0465:
#### Reference Issues/PRs
N/A

This PR ensures that warnings generated during data conversion are surfaced in the top-level validation output.

Previously, if target_column was not provided, to_sktime_format would default to using the first column and store a warning inside the adapter metadata. However, this warning was not included in validation["warnings"], which is what downstream users (including agents) are expected to check.

With this change, any warnings generated during conversion are now propagated to the top-level validation report for both synchronous and asynchronous data loading. This makes the behavior more transparent and easier to debug.

No.

- Whether metadata-level conversion warnings should be merged into top-level validation
- Whether the helper location in `executor.py` is appropriate
- Test coverage for sync and async loading

#### Any other comments?
Added regression tests for sync and async `load_data_source` warning propagation.

Validation:
- `python -m compileall src/sktime_mcp/runtime/executor.py tests/test_data_validation_warnings.py`
- Manual sync and async checks confirm the default-target warning appears in `validation["warnings"]`

Note: local `pytest` exits with code `-1` before output in this environment, so I verified behavior directly.

##### For all contributions
- [ ] Optionally, I've updated CODEOWNERS.
- [x] I've added unit tests and made sure they pass locally where possible.

For new estimators

- [x] Not applicable

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -32,6 +32,28 @@ DEMO_DATASETS = {
 }
 
 
+def _merge_adapter_validation_warnings(
+    validation_report: dict[str, Any],
+    metadata: dict[str, Any],
+) -> dict[str, Any]:
+    """Merge warnings added during adapter conversion into validation output."""
+    metadata_validation = metadata.get("validation")
+    if not isinstance(metadata_validation, dict):
+        return validation_report
+
+    metadata_warnings = metadata_validation.get("warnings", [])
+    if not metadata_warnings:
+        return validation_report
+
+    merged = validation_report.copy()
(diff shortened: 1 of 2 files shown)


Shashankss1205:
Verified on current main that the default-target warning only lives in metadata.validation.warnings; with this branch it reaches the top-level validation.warnings for both sync and async loaders. Full suite green. Merging (the new test file will be picked up by the repo-wide ruff format in the lint fix).

Shashankss1205:
Landed on main as 6a7262e (squash-merged locally because the org policy rejects the merge API for this token). Thanks!

PR #448, opened 2026-05-04
  FIX: start event loop thread when fit_predict_async called outside MCP server context

### LINKED ISSUE #446: [BUG] background_training_example.py: async job stuck at 0% progress indefinitely
**Describe the bug**
examples/background_training_example.py hangs indefinitely at 0% in Step 3. The background training job never progresses and the while loop never exits.

**To Reproduce**
```python
python examples/background_training_example.py
```

**Expected behavior**
The ARIMA training job should complete and progress should reach 100%.

**Additional context**

Image

**Versions**
<details>

  python: 3.13.11 | macOS-26.3-arm64
  sktime: 0.40.1
  sktime-mcp: 0.1.0
  pandas: 2.3.3

</details>

echo-xiao:
#### Reference Issues/PRs
Fixes #446 

- Fixed background async training job stuck at 0% progress when fit_predict_async_tool is called outside the MCP server context. The event loop exists but is not running, so scheduled coroutines never execute. Fix starts a background thread to run the event loop when it is not already running.

- Fixed KeyError: 'data_source' in background_training_example.py by supporting both 'data_source' and 'dataset' keys for compatibility between PyPI and source versions.

No. Uses only Python's built-in threading and asyncio modules.

- The threading fix in fit_predict.py: whether starting a daemon thread is the correct approach when the event loop is not running
- Whether this fix is safe to use within the MCP server context as well

#### Any other comments?
This bug was discovered while running the example scripts directly with `python examples/background_training_example.py`. The root cause is that fit_predict_async_tool uses run_coroutine_threadsafe(), which requires a running event loop. When called outside the MCP server, the loop exists but is not running, so coroutines are scheduled but never executed.

##### For all contributions
- [ ] I've added the tool to the online documentation in `docs/source/`.

### DIFF
--- a/examples/background_training_example.py
+++ b/examples/background_training_example.py
@@ -51,7 +51,7 @@ def main():
     print("✅ Training started!")
     print(f"   Job ID: {job_id}")
     print(f"   Estimator: {job_result['estimator']}")
-    print(f"   Dataset: {job_result['data_source']}")
+    print(f"   Dataset: {job_result.get('data_source') or job_result.get('dataset', 'unknown')}")
     print(f"   Horizon: {job_result['horizon']}")
 
     # Step 3: Monitor progress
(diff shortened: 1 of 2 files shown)


Shashankss1205:
It has merge conflicts in both the files, can you please fix and make a PR?

PR #451, opened 2026-05-08
FIX: raise error for invalid explicit frequency in data adapters

Saloni-0465:
#### Reference Issues/PRs
N/A

This PR fixes an issue where invalid explicit frequency values were being silently ignored in the data adapters.

Previously, if a user passed an invalid frequency like frequency="not_a_freq", the error raised by asfreq() was suppressed in both PandasAdapter and FileAdapter, and the data continued loading normally. This made it look like the requested frequency had been applied when it actually hadn’t.

With this change, both adapters now raise a clear ValueError when an invalid explicit frequency is provided. Valid frequency values continue to behave as expected.

No.

Whether failing fast is the preferred behavior for invalid explicit frequency config.
Clarity of the error message.

#### Any other comments?
Validation:
- `python -m compileall src/sktime_mcp/data/adapters/pandas_adapter.py src/sktime_mcp/data/adapters/file_adapter.py tests/test_data_frequency_validation.py`
- Manual checks for invalid pandas/file frequencies and valid daily frequency passed.

### DIFF
--- a/src/sktime_mcp/data/adapters/file_adapter.py
+++ b/src/sktime_mcp/data/adapters/file_adapter.py
@@ -100,8 +100,10 @@ class FileAdapter(DataSourceAdapter):
         # Set frequency if specified
         freq = self.config.get("frequency")
         if freq:
-            with contextlib.suppress(Exception):
+            try:
                 df = df.asfreq(freq)
+            except Exception as e:
+                raise ValueError(f"Invalid frequency '{freq}': {e}") from e
 
         self._data = df
 
--- a/src/sktime_mcp/data/adapters/pandas_adapter.py
+++ b/src/sktime_mcp/data/adapters/pandas_adapter.py
@@ -5,7 +5,6 @@ Supports loading data from pandas DataFrames with automatic
 time index detection and validation.
 """
 
-import contextlib
 from typing import Any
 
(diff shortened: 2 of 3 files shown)


Shashankss1205:
The fix is right and still needed on current main, but the branch no longer merges into a working tree: main has since added _safe_infer_freq in pandas_adapter.py, which uses contextlib, and this PR removes the `import contextlib` line, so after merging the module fails to import (31 test failures, ruff F821). Please rebase onto main and keep the import. sql_adapter.py has the same asfreq inside contextlib.suppress, so please give it the same treatment for parity. With that, ready to merge.

Me-Priyank:
Heads up on the rebase, hunk 1 drops `import contextlib`, but main's `_safe_infer_freq` still uses it at `pandas_adapter.py:24`. It merges without conflict markers so it's easy to miss; any load without an explicit frequency then raises `NameError`. The three new tests don't catch it since none of them reach `_safe_infer_freq`.

Also `except Exception` catches more than bad freq strings, a duplicate index with a valid `'D'` reports `Invalid frequency 'D': cannot reindex on an axis with duplicate labels`. Validating up front with `to_offset(freq)` and letting other failures through might be tighter.

PR #452, opened 2026-05-09
Fix csv example tmp path

### LINKED ISSUE #447: [BUG] csv_example.py and pandas_example.py: TypeError - fit_predict() got unexpected keyword argument 'data_handle'
**Describe the bug**
examples/csv_example.py and examples/pandas_example.py both fail with a TypeError when calling fit_predict() with the data_handle argument. The README documents data_handle as a supported parameter, but the Executor.fit_predict() method does not implement it in the PyPI release.

**To Reproduce**

```python
pip install sktime-mcp
python examples/csv_example.py
python examples/pandas_example.py
```

**Expected behavior**
fit_predict() should accept a data_handle returned from load_data_source() and use it as input data for forecasting.

**Additional context**
Error output:
TypeError: Executor.fit_predict() got an unexpected keyword argument 'data_handle'

Affects both:
- examples/csv_example.py
- examples/pandas_example.py

The README documents this usage:
{
  "estimator_handle": "est_abc123",
  "data_handle": "data_abc123",
  "horizon": 12
}

Root cause: This error occurs with the PyPI-installed version (0.1.0) but does NOT reproduce with an editable install from source (pip install -e .). This suggests the PyPI release is out of sync with the latest codebase.

- ❌ Fails: /opt/miniconda3/bin/python (PyPI install 0.1.0)
- ✅ Works: .venv/bin/python (editable install from source)

Image
Image

**Versions**
<details>

python: 3.13.11 | macOS-26.3-arm64
sktime: 0.40.1
sktime-mcp: 0.1.0 (PyPI)
pandas: 2.3.3

</details>

Nischal1425:
Fixes #447 

`csv_example.py` used hardcoded `/tmp/` paths which don't exist on Windows, causing an `OSError`. Replaced with `tempfile.gettempdir()` which works cross-platform.

### DIFF
--- a/examples/csv_example.py
+++ b/examples/csv_example.py
@@ -5,6 +5,7 @@ This example demonstrates how to use the sktime-mcp data loading
 functionality with CSV files.
 """
 
+import tempfile
 from pathlib import Path
 
 import pandas as pd
@@ -20,7 +21,7 @@ sample_data = pd.DataFrame(
     }
 )
 
-csv_path = Path("/tmp/sample_sales_data.csv")
+csv_path = Path(tempfile.gettempdir()) / "sample_sales_data.csv"
 sample_data.to_csv(csv_path, index=False)
 print(f"Created sample CSV file: {csv_path}")
 print(f"File size: {csv_path.stat().st_size} bytes")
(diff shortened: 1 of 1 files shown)


Shashankss1205 (commented):
This is actually stale because we have removed all examples because of the new updated docs

PR #453, opened 2026-05-11
initial tests for docker setup

### LINKED ISSUE #444: Working on a Docker hosted sktime-mcp server

Shashankss1205:
#### Reference Issues/PRs

Closes #444 . Adds Docker containerization support for sktime-mcp.

Adds end-to-end Docker support for running sktime-mcp without any local Python setup. This includes:

**New files:**
- **`Dockerfile`** - Multi-stage build (builder → slim runtime). The builder stage installs build tools (`build-essential`, `gfortran`, `libopenblas-dev`) and all Python dependencies, then only the installed packages are copied into a minimal `python:3.12-slim` runtime image. Runs as a non-root `mcp` user for security.
- **`.dockerignore`** - Filters out `.git`, `__pycache__`, `.venv`, `docs/_build`, and other dev artifacts to keep the build context small (~456KB).
- **`docker-compose.yml`** - Service definition with configurable environment variables (`SKTIME_MCP_LOG_LEVEL`, `SKTIME_MCP_AUTO_FORMAT`, etc.) and a named `sktime-models` volume for persisting saved models across container restarts.

**Modified files:**
- **`README.md`** - Added a 🐳 Docker section under Installation with build/run/compose commands and a Claude Desktop JSON config example for Docker-based MCP. Updated the project structure tree to include the new files.
- **`Makefile`** - Added `make docker-build` and `make docker-run` convenience targets.

No new Python dependencies. Docker is the only external requirement, and it is optional - existing pip-based installation is unaffected.

- The multi-stage Dockerfile approach - builder installs `build-essential`/`gfortran`/`libopenblas-dev` for compiling numpy/scipy, but only `libopenblas0` is carried into the runtime image. Is this the right trade-off vs. a single-stage build?
- The `docker-compose.yml` uses `stdin_open: true` with `tty: false` - this is intentional since MCP JSON-RPC needs raw stdin, not a pseudo-TTY. Worth confirming this works across Docker versions.
- Claude Desktop config in README uses `"command": "docker"` with `"args": ["run", "-i", "--rm", "sktime-mcp"]` - reviewers familiar with MCP client setups should verify this pattern.

#### Any other comments?

Verified end-to-end locally:
- ✅ `docker build` completes successfully (~81s)
- ✅ All package imports work inside the container (`sktime-mcp v0.1.0`, `sktime`, `mcp`)
- ✅ MCP `initialize` JSON-RPC handshake succeeds
- ✅ `tools/list` returns all 22 registered tools

##### For all contributions
- [ ] I've added unit tests and made sure they pass locally (`make check`).
- [ ] I've added the tool to the online documentation in `docs/source/`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,65 @@
+# =============================================================================
+# Dockerfile for sktime-mcp
+#
+# Multi-stage build:
+#   1. Builder  – installs build tools and Python dependencies
+#   2. Runtime  – slim image with only the installed packages
+#
+# Usage:
+#   docker build -t sktime-mcp .
+#   docker run -i sktime-mcp                  # stdio transport (default)
+#   docker run -i -e SKTIME_MCP_LOG_LEVEL=DEBUG sktime-mcp
+# =============================================================================
+
+# ---------------------------------------------------------------------------
+# Stage 1: Builder
+# ---------------------------------------------------------------------------
+FROM python:3.12-slim AS builder
+
+# Avoid interactive prompts during package install
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Install minimal build dependencies required by some scientific packages
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        build-essential \
+        gfortran \
+        libopenblas-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /build
+
+# Copy only dependency-related files first to maximise Docker layer caching
+COPY pyproject.toml README.md LICENSE ./
+COPY src/ src/
+
+# Install the project in production mode (no dev/test extras)
+RUN pip install --no-cache-dir --prefix=/install .
+
+# ---------------------------------------------------------------------------
+# Stage 2: Runtime
+# ---------------------------------------------------------------------------
+FROM python:3.12-slim AS runtime
+
+# Runtime library needed by numpy/scipy linked against OpenBLAS
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        libopenblas0 \
+    && rm -rf /var/lib/apt/lists/*
+
+# Copy installed Python packages from the builder stage
+COPY --from=builder /install /usr/local
+
+# Create a non-root user for security
+RUN useradd --create-home --shell /bin/bash mcp
+USER mcp
+WORKDIR /home/mcp
+
+# Environment defaults (overridable at runtime)
+ENV SKTIME_MCP_LOG_LEVEL=WARNING
+ENV PYTHONUNBUFFERED=1
+
+# Health-check: verify the package is importable
+HEALTHCHECK --interval=30s --timeout=5s --retries=3 \
+    CMD python -c "import sktime_mcp; print('ok')" || exit 1
+
+# MCP servers communicate over stdio, so the entrypoint is the CLI command
+ENTRYPOINT ["sktime-mcp"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,7 +19,7 @@ services:
     image: sktime-mcp:latest
     container_name: sktime-mcp
 
-    # stdin_open + tty are required for stdio-based MCP transport
+    # Keep stdin open, but avoid allocating a pseudo-TTY so MCP gets raw JSON-RPC over stdio.
     stdin_open: true
     tty: false  # MCP JSON-RPC needs raw stdin, not a pseudo-TTY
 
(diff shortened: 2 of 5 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,38 @@
+# =============================================================================
+# Docker Compose for sktime-mcp
+#
+# Usage:
+#   docker compose build          # build the image
+#   docker compose run sktime-mcp # run with stdio (interactive, for MCP clients)
+#
+# Notes:
+#   - MCP uses stdio transport, so `docker compose run` (with TTY) is the
+#     correct invocation rather than `docker compose up` (which is for daemons).
```
geetu040: we disable tty, so a better wording would be this, right?
```suggestion
#   - MCP uses stdio transport, so `docker compose run` with stdin attached is the
#     correct invocation rather than `docker compose up` (which is for daemons).
```

docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+
+services:
+  sktime-mcp:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    image: sktime-mcp:latest
+    container_name: sktime-mcp
+
+    # stdin_open + tty are required for stdio-based MCP transport
```
geetu040: as said above
```suggestion
    # Keep stdin open, but avoid allocating a pseudo-TTY so MCP gets raw JSON-RPC over stdio.
```

docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    environment:
+      - SKTIME_MCP_LOG_LEVEL=${SKTIME_MCP_LOG_LEVEL:-WARNING}
+      - SKTIME_MCP_AUTO_FORMAT=${SKTIME_MCP_AUTO_FORMAT:-true}
+      - SKTIME_MCP_JOB_MAX_AGE_HOURS=${SKTIME_MCP_JOB_MAX_AGE_HOURS:-24}
+      - SKTIME_MCP_JOB_CLEANUP_INTERVAL=${SKTIME_MCP_JOB_CLEANUP_INTERVAL:-3600}
+
+    volumes:
+      # Persist saved models outside the container
+      - sktime-models:/home/mcp/models
```
geetu040, line 34: I tried running this through Codex as an MCP server:

```bash
codex mcp add sktime -- docker run -i --rm -v sktime-models:/home/mcp/models sktime-mcp
```

Then I asked it to instantiate a `NaiveForecaster`, run `fit_predict` on the built-in `airline` dataset, and save the fitted model to `/home/mcp/models/my-dummy-model`.

The MCP flow worked until `save_model`, which failed with:

```text
Called sktime.save_model({"estimator_handle":"est_2c26c61bbe5b","path":"/home/mcp/models/my-dummy-model"})
→ {"success": false, "error": "`mlflow` is an extra dependency and is not included in the base sktime installation. Please run `pip install mlflow` or `pip install sktime[mlflow]` to install the package.To install the requirement 'mlflow' or 'mlflow-skinny', please run: `pip install mlflow` or `pip install mlflow-skinny` "}
```

In general, I think it’s fine for optional MCP tools to fail gracefully when an extra dependency is not installed and that leaves us with the bigger question of whether we should include extra deps. But in this Docker setup we expose a `/home/mcp/models` volume specifically for persisted models, so I’d expect the image to include the dependency needed for that workflow.

Let's discuss this over the next comment.

Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /build
+
+# Copy only dependency-related files first to maximise Docker layer caching
+COPY pyproject.toml README.md LICENSE ./
+COPY src/ src/
+
+# Install the project in production mode (no dev/test extras)
+RUN pip install --no-cache-dir --prefix=/install .
```
geetu040, line 36: so the dockerized mcp server basically cannot run anything that requires an optional dep?

geetu040 (changes requested):
Overall the changes look goods. I have set this up locally and used this mcp server through docker container in codex, and is working fine.

I have left some comments, which are not hard-blockers, rather nits and some require discussion.

Dockerfile
Shashankss1205: I created another issue https://github.com/sktime/sktime-mcp/issues/475 to address the same problem and needs some design decisions. I think we can merge this PR by then for simpler estimators.

geetu040 (approved):
Looks good in the current state. We can discuss over the deps in #483.

PR #463, opened 2026-05-30
Remove legacy discovery tools list_estimators, search_estimators in favour of new tools query_registry, describe_component

### LINKED ISSUE #461: Phase 1: Discovery & Registry Tools
The goal of this phase is to cleanly isolate and refactor the **Discovery Tools**. These tools inspect the native `sktime` registry and **must require no active handles** (i.e., no instantiated objects or state). This provides a stateless, performant, and intuitive interface for LLM clients to search, filter, and inspect the available algorithms in `sktime` before instantiating them.

### 1. `list_estimators`
Searches and filters the native `sktime` estimator registry. Returns basic metadata, import path, and capability tags for matching classes.

**Arguments:**
* `task` *(str, optional)*: Task filter (e.g., `"forecasting"`, `"classification"`).
* `tags` *(dict[str, Any], optional)*: Tag filters (e.g., `{"capability:pred_int": true}`).
* `query` *(str, optional)*: Text search substring for matching estimator names.
* `limit` *(int, optional, default=50)*: Maximum number of results to return (for pagination).
* `offset` *(int, optional, default=0)*: Offset for pagination.

**Returns:**
* `dict`: `{"estimators": list[dict], "total": int}`

### 2. `describe_estimator`
Retrieves full docstrings, hyperparameter descriptions, default values, capability tags, and the Python import path for a specific named class.

**Arguments:**
* `estimator` *(str, required)*: Exact name of the estimator class (e.g., `"ARIMA"`).

**Returns:**
* `dict`: `{"name": str, "description": str, "params": dict, "tags": dict, "import_path": str}`

### 3. `get_available_tags`
Returns a dictionary of all queryable capability tags across the `sktime` registry with their descriptions and expected value types, helping clients understand what they can filter on.

**Arguments:**
* *None*

**Returns:**
* `dict`: `{"tags": dict[str, dict]}`

## Implementation Notes
* **`list_estimators`**: Utilize `sktime.registry.all_estimators()`. Implement the custom `query` filtering and pagination (`limit`/`offset`) logic on top of the returned results.
* **`describe_estimator`**: Dynamically locate the class via the registry, then use `inspect` (or sktime's internal param/doc retrieval utilities) to extract the docstring, `get_class_tags()`, and `get_param_names()`.
* **`get_available_tags`**: Wrap `sktime.registry.all_tags()`.

## Acceptance Criteria
- [ ] Implement the `list_estimators`, `describe_estimator`, and `get_available_tags` tools as defined above.
- [ ] Ensure that none of these tools require active handles (purely stateless operation).
- [ ] Define precise JSON schemas (via Pydantic or equivalent) for the tool arguments and expected outputs.
- [ ] Add unit tests verifying pagination, filtering (tags/tasks), and edge cases (e.g., requesting a non-existent estimator).
- [ ] Remove any deprecated discovery tool endpoints that conflict with this new structure.

Shashankss1205:
#### Reference Issues/PRs

Fixes #461

**Independence:** This PR is rebased on latest `main` and does **not** depend on any other open PR. It can be merged on its own.

This pull request removes the legacy discovery tools (`list_estimators`, `search_estimators`, `describe_estimator`, and `get_available_tags`) and routes discovery through unified replacements:

- **`query_registry`** — unified registry search (estimators, tags, metrics)
- **`describe_component`** — detailed info for any sktime component class

**Changes:**
- Removed legacy tool registration and routing from `server.py`
- Renamed `describe_estimator.py` → `describe_component.py`
- Updated `list_estimators.py` internals used by `query_registry`
- Updated tests in `tests/test_core.py` and `tests/test_discovery_consolidated.py`
- Updated docs in `README.md`, `docs/source/implementation.md`, `docs/source/usage-examples.md`

No.

- Verify `query_registry` and `describe_component` work for estimators, tags, and pagination
- Confirm no legacy discovery tool names remain in production routing paths

#### Any other comments?

Rebased on latest `main`. `make check` passes locally.

##### For all contributions


fkiraly (changes requested):
Overall, this looks like an improvement.

However, can you please make sure to include docstrings to all important functions and classes? Please have a look how `sktime` does this. Preferably in the same `numpydoc` adjacent style that `sktime` uses.

Otherwise, it becomes really hard to review changes, especially if they are subtle changes.

E.g., I have the question what the new `target` variable means and which values it can assume, but there is no piece of documentation that would tell me.

Shashankss1205:
@fkiraly thanks for the feedback, I will try updating all of the PRs with the suitable doc strings to let the maintainers understand the pieces of code easily...

Shashankss1205:
Heyy @fkiraly can you please check the latest state of this PR?

fkiraly (changes requested):
* I thought we discussed (and agreed?) to deduplicate `target` and `task`, at least not making "metrics" separate from "estimators"?
* there is a more severe root problem: the module `registry.interface` is largely a duplicate of `sktime.registry`. Why is there a need to wrap all `sktime` functions in 1:1 duplicates and rename tasks via a dict, e.g., `"forecaster"` to `"forecasting"` etc?
    * use the `sktime` registry mechanisms where possible. Do not duplicate and multiply the maintenance surface

Some examples in `RegistryInterface`:

* remove `TASK_MAP`, do you want to maintain minimally different names for every single `sktime` task which ahve standardized task names?
* `_get_tags` duplicates ... `get_tags`. Also, the `_tags` attribute should never be directly accessed.
* `_get_hyperparameters` duplicates `get_params`, `get_param_names`. `get_param_defaults`.
* `get_all_estimators` duplicates logic in `all_estimators`.
* _filter_by_tags` duplicates logic in `all_estimators`
* `get_estimator_by_name` duplicates `registry.craft`
* `get_available_tags` duplicates the tag registry
* registry caching by singleton is duplicative, because the first `all_estimators` call is already cached.

fkiraly:
I would suggest: deduplicate as much as possible. Avoid AI hallucinated boilerplate duplication, instead drive logic and also value conventions (such as `object_type` values like `"forecasting"` for estimator type), much closer to using available `sktime` methods.

Deduplication will also suggest a better signature for the "query" method (e.g., not `target` and `task`, but terminology also used in `sktime` methods).
I do not think it is better for LLM that have been trained on lots of `sktime` material to suddenly redefine terminology or the unique strings for task types.

fkiraly (approved):
Looks good, I think most requests are addressed.

I would suggest, let´s unblock the stack and then have another round of reviews once it is complete and works.

PR #464, opened 2026-06-01
Use training dataset for export_code fit example

### LINKED ISSUE #281: [ENH] `export_code` with `include_fit_example=True` silently falls back to 'airline' dataset

TheDerbiedOne:
Fixes #281.

- Record the demo dataset used in `fit_predict` / `fit_predict_async` on the handle metadata as `training_dataset`.
- When `export_code(..., include_fit_example=True)` is called with no explicit `dataset`, default to `training_dataset` before falling back to `airline`.
- Add a focused unit test for the new default behavior.

Notes: I could not run the test suite locally in this environment because only Python 3.9 is available here and the project uses Python 3.10+ type syntax; the changes are small and covered by an updated unit test in CI.

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -277,6 +277,16 @@ class Executor:
         if not fit_result["success"]:
             return fit_result
 
+        # Record the training data source on the handle for later use (e.g. export_code defaults).
+        try:
+            handle_info = self._handle_manager.get_info(handle_id)
+            if dataset:
+                handle_info.metadata["training_dataset"] = dataset
+            elif data_handle:
+                handle_info.metadata["training_data_handle"] = data_handle
+        except Exception as e:  # pragma: no cover
+            logger.debug("Could not record training source metadata: %s", e)
+
         return self.predict(handle_id, fh=fh, X=X)
 
     async def fit_predict_async(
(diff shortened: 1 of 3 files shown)


Shashankss1205:
Duplicated as https://github.com/sktime/sktime-mcp/pull/286

Issue #478, opened 2026-06-12
Phase 3: Execution & Application Tools


Shashankss1205:
The goal of this phase is to cleanly isolate and refactor the **Execution & Application Tools**. These tools apply instantiated estimator handles to data — fitting, predicting, transforming, and updating models. This phase standardizes how handles interact with datasets (via `data_handle` or built-in demo names) and ensures seamless support for both synchronous and async/background execution.

Part of the umbrella refactoring tracked in #460. Design reference: #270.

### 1. `fit` / `fit_async`
Fit an instantiated estimator or pipeline on target series `y` and optional exogenous data `X`.

**Arguments:**
* `estimator_handle` *(str, required)*: Active estimator handle.
* `y` *(str, required)*: Target time series — active `data_handle` or built-in demo dataset name (e.g., `"airline"`).
* `X` *(str, optional)*: Exogenous series — `data_handle` or demo name.
* `fit_params` *(dict[str, Any], optional)*: Extra parameters forwarded to the underlying `fit` call (e.g., `fh` if required at fit time).

**Returns (`fit`):**
* `dict`: `{"success": bool, "estimator_handle": str, "message": str}`

**Returns (`fit_async`):**
* `dict`: `{"success": bool, "job_id": str, "message": str}`

---

### 2. `predict` / `predict_async`
Generate predictions from a fitted estimator or pipeline.

**Arguments:**
* `estimator_handle` *(str, required)*: Fitted estimator handle.
* `X` *(str, optional)*: Exogenous/feature series (`data_handle` or demo name).
* `fh` *(list[int] | list[str] | int, optional)*: Forecast horizon (for forecasters).
* `predict_params` *(dict[str, Any], optional)*: Extra parameters forwarded to `predict`.

**Returns (`predict`):**
* `dict`: `{"success": bool, "predictions": list[dict], "fh": list}`

**Returns (`predict_async`):**
* `dict`: `{"success": bool, "job_id": str, "message": str}`

---

### 3. `predict_interval`
Generate prediction intervals from a fitted probabilistic forecaster.

**Arguments:**
* `estimator_handle` *(str, required)*: Fitted estimator handle.
* `X` *(str, optional)*: Exogenous/feature series.
* `fh` *(list[int] | list[str] | int, optional)*: Forecast horizon.
* `coverage` *(list[float] | float, optional, default=0.9)*: Nominal coverage interval(s) in `(0, 1)`.

**Returns:**
[...]

PR #488, opened 2026-06-29
feat: Add full scitype routing and dynamic call_method tool

Shashankss1205:
### The PR includes the following features:
1. **Dynamic Method Reflection (`call_method`)**: A new universal tool that allows agents to interact with non-standard scitypes (like `Metric`, `Splitter`, `Aligner`) by invoking their native methods (e.g., `.split()`, `.get_alignment()`) dynamically.
2. **Intelligent Data Injector**: Dynamic `kwargs` replacement in `call_method` (e.g. `y_dataset: airline`), automatically loading datasets into memory and mapping them directly to method arguments before execution.
3. **Multi-Scitype Argument Routing**: Intelligent routing inside `executor.fit` and `executor.predict` to dispatch correct positional arguments (like `X` vs `y`) based on whether the object is a Forecaster, Classifier, Regressor, Clusterer, or Transformer.
4. **Instructional Guardrails**: Clean guardrails that catch invalid `fit()` and `predict()` calls on objects that don't support them, returning specific instructional payloads that teach AI agents to pivot to the `call_method` tool instead of crashing.

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -117,29 +117,31 @@ class Executor:
 
     def instantiate(
         self,
-        estimator_name: str,
-        params: dict[str, Any] | None = None,
+        spec: str,
     ) -> dict[str, Any]:
-        """Instantiate an estimator and return a handle."""
-        node = self._registry.get_estimator_by_name(estimator_name)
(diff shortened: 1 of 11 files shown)


Shashankss1205:
Fixes #478

Issue #481, opened 2026-06-12
Phase 7: Background Jobs


Shashankss1205:
The goal of this phase is to cleanly isolate and refactor the **Background Jobs** tool group. These tools provide standardized polling, tracking, and management of asynchronous task executions — enabling LLM clients to kick off long-running fit, predict, and evaluate operations without blocking the conversation.

Part of the umbrella refactoring tracked in #460. Design reference: #270.

### 1. `check_job_status`
Inspect the status, execution progress, and results of a background job.

**Arguments:**
* `job_id` *(str, required)*: Unique ID of the job (returned by `*_async` execution tools).

**Returns:**
* `dict`: `{"success": bool, "job_id": str, "status": str, "progress_pct": float, "current_step": str, "error": str|null, "result": dict|null}`

Valid `status` values: `"pending"`, `"running"`, `"completed"`, `"failed"`, `"cancelled"`.

---

### 2. `list_jobs`
List active and historical background jobs with optional filtering and pagination.

**Arguments:**
* `status` *(str, optional)*: Filter by status (`"pending"`, `"running"`, `"completed"`, `"failed"`, `"cancelled"`).
* `limit` *(int, optional, default=20)*: Maximum number of jobs to return.
* `offset` *(int, optional, default=0)*: Pagination offset.

**Returns:**
* `dict`: `{"success": bool, "jobs": list[dict], "total": int}`

---

### 3. `cancel_job`
Request cancellation of a running or pending background job.

**Arguments:**
* `job_id` *(str, required)*: Unique ID of the job to cancel.

**Returns:**
* `dict`: `{"success": bool, "job_id": str, "message": str}`

## Implementation Notes
* Centralize job lifecycle management in the runtime job manager (`sktime_mcp.runtime.jobs`).
* Async execution tools from Phases 3 and 4 (`fit_async`, `predict_async`, `evaluate_async`, `load_data_source_async`) must all register jobs through the same interface.
* Implement automatic cleanup of stale jobs based on `SKTIME_MCP_JOB_MAX_AGE_HOURS` and `SKTIME_MCP_JOB_CLEANUP_INTERVAL` settings.
* Include meaningful `current_step` progress messages (e.g., `"Fitting ARIMA on airline..."`) for LLM-friendly status updates.
* Sanitize job results before returning to ensure JSON serializability.

## Acceptance Criteria
- [ ] Implement `check_job_status`, `list_jobs`, and `cancel_job` with signatures matching the specification above.
[...]

PR #489, opened 2026-06-29
feat: Restore async background training to mcp_sktime_fit

Shashankss1205:
#### Reference Issues/PRs

Follow-up to #487 (Restores missing async background training).

When the legacy monolithic `fit_predict_async_tool` was removed in PR #487 during the atomic tools refactoring, the ability to run estimator training asynchronously in the background was accidentally lost. 

This PR cleanly restores background training directly into the atomic `mcp_sktime_fit` tool. 

**Changes:**
1. Added a `run_async` boolean flag to the `fit` tool's input schema in `server.py`.
2. Updated `fit_predict.py` to route async calls to a new background job.
3. Implemented `fit_async` inside `executor.py` utilizing `asyncio.run_in_executor` alongside thread pools to ensure the main MCP server event loop is not blocked during hours-long model training.
4. Agents can now pass `{"run_async": true}` and immediately receive a `job_id` to poll using the `check_job_status` tool.

No new dependencies introduced. Uses the existing `asyncio` and `concurrent.futures.ThreadPoolExecutor` from the standard library.

- [x] Concurrency logic in `executor.fit_async()`: Validating that using a `ThreadPoolExecutor` correctly prevents the `fit` method from blocking the primary async server loop.
- [x] Ensure the `run_async` parameter schema in `server.py` is perfectly aligned with `load_data_source` for consistency across tools.

#### Any other comments?

This allows advanced models like deep learning networks or large parameter grid searches to run indefinitely in the background without causing the MCP client to timeout!

##### For all contributions

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -514,6 +514,94 @@ class Executor:
 
         return self.predict(handle_id, fh=fh, X=X)
 
+    async def fit_async(
+        self,
+        handle_id: str,
+        dataset: str | None = None,
+        data_handle: str | None = None,
+        job_id: str | None = None,
+    ) -> dict[str, Any]:
+        """Async version of fit with job tracking."""
+        try:
+            import asyncio
+            from sktime_mcp.runtime.jobs import JobStatus
+            
+            # Update status to RUNNING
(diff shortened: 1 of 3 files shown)


Shashankss1205:
Fixes #481

PR #514, opened 2026-07-19
fix: support multi-series X for aligner fit

### LINKED ISSUE #491: [BUG]The Aligner Limitation (Nested Panel Data)
When I instantiated AlignerDTW and tried to call mcp_sktime_fit(dataset="airline"), the MCP server returned exactly this:

```json
{
  "success": false,
  "error": "{'df-list': \"X must be of python type list, or a subtype thereof, but found <class 'pandas.core.series.Series'>\"}"
}
```

Limitation: Aligners mathematically align multiple time series, so their native fit() method strictly requires a Python list of DataFrames/Series (e.g. X=[df1, df2]). The MCP tool mcp_sktime_fit accepts a single dataset parameter and loads a single DataFrame. There is currently no schema design in the tools for an LLM to say "Load these 5 datasets and pass them as a single list to X".

Saswatsusmoy:
#### Reference Issues/PRs

Fixes #491

Aligners need `X` as a panel / df-list (list of series). `fit` only resolved a single series per `X_*` / `y_*` slot and routed `object_type=aligner` through the forecaster-style `fit(y, X=...)` path. That produced the df-list type error on a single series (e.g. airline) and a signature clash when only `X` was set.

This change:

- lets `X_handle` / `X_dataset` be a string or a list of ids
- assembles a df-list for list inputs (Series → one-column DataFrame)
- routes aligners to `fit(X)` with a clear error on a single series/table
- extends `call_method` `*_dataset` / `*_data_handle` injection the same way for multi-series lists

Shared resolve is used by sync `fit` and `fit_async`. Tests use `AlignerNaive` (no soft deps).

No.

- List resolve for `X_handle` / `X_dataset` and Series → DataFrame for df-list items
- Aligner branch in `Executor.fit` (including single flat DataFrame vs MultiIndex panel)
- `call_method` list injection parity with fit (`len >= 2`)

#### Any other comments?

`make check` is green locally (206 tests). Commit-time pre-commit passed on this branch.

##### For all contributions
- [ ] I've added the tool to the online documentation in `docs/source/`. n/a (schema/description only on the existing `fit` tool; no new tool)
- [ ] I've updated the existing example scripts or provided a new one to showcase how my tool works in `examples/`. n/a

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -55,6 +55,34 @@ def _get_demo_datasets() -> dict:
     return _DEMO_DATASETS
 
 
+def _as_frame(obj: Any) -> Any:
+    """Series → one-column DataFrame for df-list items."""
+    if isinstance(obj, pd.Series):
+        return obj.to_frame()
+    return obj
+
+
+def _is_single_series_table(obj: Any) -> bool:
+    """True for a lone Series or flat DataFrame (not a MultiIndex panel)."""
+    if isinstance(obj, pd.Series):
+        return True
+    return isinstance(obj, pd.DataFrame) and not isinstance(obj.index, pd.MultiIndex)
+
+
+def _dataset_slot(data_res: dict[str, Any], slot: str) -> Any:
+    """Pick y or X from a load_dataset result. X falls back to y when absent."""
+    if slot == "X":
+        return data_res["X"] if data_res["X"] is not None else data_res["y"]
+    return data_res["y"]
+
+
+_ALIGNER_NEED_LIST_MSG = (
+    "Aligners require multiple series as X, not a single series/table. "
+    "Pass X_handle or X_dataset as a list of at least two ids."
+)
+_X_LIST_MIN_MSG = "X as a list needs at least two series."
+
+
 def _to_period_index_if_possible(obj: Any) -> Any:
     """Return *obj* with a ``PeriodIndex`` when its index is a regular datetime index.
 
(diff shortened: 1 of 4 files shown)


Saswatsusmoy:
one caveat for reviewers / users: a single series (e.g. y_dataset=airline alone) still fails, on purpose. aligners need multiple series, so we return a short error asking for X_handle or X_dataset as a list of at least two ids. the bug was never being able to pass that list and wrong fit routing. multi-series list path is what this pr fixes.

Shashankss1205:
Verified the failure on current main (both the BaseAligner.fit() signature clash and the df-list mtype error) and that this branch fixes it via list-valued X_handle/X_dataset without adding a tool; full suite green with it, ruff clean. It now conflicts with #515 (merged first) in exactly one hunk of sanitize_for_json in server.py: keep the _seen-threading lines from main and your `list | tuple` style. Please rebase and I will merge immediately. Note for the record: with run_async=true an unknown handle now surfaces as a failed job rather than an immediate error, which is acceptable since fit_async validates.

Saswatsusmoy:
Rebased onto current `main`.

`sanitize_for_json` keeps the `_seen` threading from #515 and the `list | tuple` style from this PR. The rest of the aligner list-X path is unchanged. Ready on your side.

Me-Priyank:
Tried this locally, applies clean and the aligner dispatch reads well.

One thing that might matter for the v1.0 contract later: `X_handle`/`X_dataset` use `anyOf`, but every other union in the schemas uses the flat `{"type": ["string", "array"]}` form, and `test_schema_types.py:47-58` pins that. Keeping a top-level `type` also avoids the stringification problem that file's docstring describes.

Also noticed the `len(v) < 2` check fires for any `*_data_handle` key in `call_method`, so `{"y_data_handle": ["h1"]}` comes back with the X-specific message.

Saswatsusmoy:
good catch, thanks. switched `X_handle`/`X_dataset` to the flat `type: ["string", "array"]` form like the other unions, and scoped the `call_method` list inject so only `X_*` gets the multi-series path. `y_data_handle: ["h1"]` no longer returns the X min-length message.

PR #515, opened 2026-07-22
fix: guard sanitize_for_json against circular references

### LINKED ISSUE #189: [BUG] sanitize_for_json crashes on circular references (RecursionError)
### Description

`sanitize_for_json` in `server.py:74-83` recursively walks objects but has no cycle detection. If an sktime estimator contains circular references in its attributes (some do via parent/child pointers), this causes `RecursionError` and crashes the MCP server.

```python
def sanitize_for_json(obj):
    if isinstance(obj, dict):
        return {str(k): sanitize_for_json(v) for k, v in obj.items()}  # no cycle check
    elif isinstance(obj, (list, tuple)):
        return [sanitize_for_json(item) for item in obj]  # no cycle check
```

The `hasattr(obj, "__dict__")` fallback on line 81 catches most non-dict/list objects by converting to `str()`, but dict/list cycles are uncaught.

### How to reproduce

Any tool returning a result containing a circular reference will trigger this. Likely scenario: an estimator with a `parent` attribute pointing back to a container.

### Suggested fix

Track visited objects by `id()`:

```python
def sanitize_for_json(obj, _seen=None):
    if _seen is None:
        _seen = set()
    obj_id = id(obj)
    if obj_id in _seen:
        return str(obj)
    _seen.add(obj_id)
    if isinstance(obj, dict):
        return {str(k): sanitize_for_json(v, _seen) for k, v in obj.items()}
    ...
```

Alternatively, wrap the top-level call in `try/except RecursionError`.

Found while reading the codebase for PR #114 work.

SarthakB11:
Closes #189.

## Problem

`sanitize_for_json` recurses into dicts/lists/tuples/Series/DataFrames
with no cycle detection. An object graph containing a back-reference
(e.g. a dict that (indirectly) contains itself) blows the recursion
limit and crashes with `RecursionError` instead of returning any
usable response to the calling tool.

## Fix

Track the `id()` of containers visited along the current recursion
path in a `_seen` set (passed as an internal-only kwarg, not part of
the public signature). On revisiting an already-seen container id,
return the placeholder string `"<circular reference>"` instead of
recursing again. Each recursive branch gets its own copy of `_seen`
(immutable set union), so sibling branches that legitimately share the
same nested object are not falsely flagged as circular.

## Tests

Added 4 regression tests to `tests/test_sanitize.py`:
- direct self-reference in a dict
- direct self-reference in a list
- indirect two-hop cycle (`a -> b -> a`)
- sibling containers sharing the same nested object are NOT flagged as circular (guards against an overly aggressive fix)

Full suite passes locally: 201 passed.

### DIFF
--- a/src/sktime_mcp/server.py
+++ b/src/sktime_mcp/server.py
@@ -143,7 +143,7 @@ def _apply_response_token_limit(tool_name: str, text: str) -> str:
     return text[:budget] + notice
 
 
-def sanitize_for_json(obj):
+def sanitize_for_json(obj, _seen=None):
     """Recursively convert objects to JSON-serializable format.
 
     Handles:
@@ -151,7 +151,19 @@ def sanitize_for_json(obj):
     - NumPy integer/float scalars and ndarrays
     - Pandas Timestamp, NaT, NA, and Series/DataFrame
     - Arbitrary objects (fallback to str repr)
+    - Circular references (returns a placeholder instead of recursing forever)
     """
+    if _seen is None:
+        _seen = set()
+
+    if isinstance(obj, (dict, list, tuple)) or (
+        _PANDAS_AVAILABLE and isinstance(obj, (pd.Series, pd.DataFrame))
+    ):
+        obj_id = id(obj)
+        if obj_id in _seen:
+            return "<circular reference>"
+        _seen = _seen | {obj_id}
+
     # --- NumPy types ---
     if _NUMPY_AVAILABLE:
         if isinstance(obj, np.integer):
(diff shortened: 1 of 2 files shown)


Shashankss1205:
Verified: a self-referencing dict/list raises RecursionError in sanitize_for_json on current main and the new tests pass here; path-scoped _seen keeps shared sub-objects serialisable. Full suite green, ruff clean. Merging; closes #189.

Shashankss1205:
Landed on main as a26e3d1 (squash-merged locally because the org policy rejects the merge API for this token). Thanks!

PR #517, opened 2026-08-06
fix: declare types for 5 untyped inputSchema properties

Shashankss1205:
## Problem

Five `inputSchema` properties in `server.py` declared no `type`/`enum`/`anyOf`/`oneOf`, so MCP clients serialize their values as strings in transit. Every explicit use of these parameters then fails Python-side type checks:

| Param | Observed error |
|---|---|
| `fit.fh` | `Invalid fh. ... found type <class 'str'>, values = 12` |
| `predict.coverage` | `'<' not supported between instances of 'int' and 'str'` |
| `predict.alpha` | same |
| `split_data.fh` | `fh must be an integer or list of integers, got str` |
| `plot_series.markers` | `There must be one marker for each time series` |

Blast radius: `predict_interval`/`predict_quantiles` were usable only with defaults, every `fh` code path in `split_data` was dead, estimators requiring `fh` at fit time could not be fitted through the server at all, and multi-series markers could not be set. (BUG-01 from the 2026-07-26 ~400-case test sweep; found independently by 4 agents.)

## Fix

Give each property an explicit union type matching what the Python side accepts, e.g.:

```python
"fh": {
    "description": "...",
    "type": ["integer", "array"],
    "items": {"type": "integer"},
},
```

## Regression guard

New `tests/test_schema_types.py`:
- walks every property of every tool's `inputSchema` (recursively) and asserts each carries a type constraint — this bug class is invisible to unit tests that call the tool functions directly, it only appears over the wire
- pins the expected types of the five previously-broken properties

The lint test confirms these five were the only untyped properties in the schema set.

## Testing

- `pytest tests/` — 208 passed

### DIFF
--- a/src/sktime_mcp/server.py
+++ b/src/sktime_mcp/server.py
@@ -340,6 +340,8 @@ async def list_tools() -> list[Tool]:
                     },
                     "fh": {
                         "description": "Optional: Forecast horizon (e.g. 12 or [1,2,3]) to pass to fit",
+                        "type": ["integer", "array"],
+                        "items": {"type": "integer"},
                     },
                     "run_async": {
                         "type": "boolean",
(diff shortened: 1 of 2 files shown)


Shashankss1205:
Merged into main as 5558f4a (local squash-merge; the API merge endpoint is blocked by the org's OAuth app restrictions). Closing.

PR #518, opened 2026-08-06
fix: escape string values in export_code generated code

Shashankss1205:
## Problem

`export_code` emitted syntactically invalid Python for any spec containing a string argument (BUG-02 from the 2026-07-26 test sweep). `_format_value` wrapped strings in bare double quotes with no escaping:

```python
if isinstance(value, str):
    return f'"{value}"'
```

so exporting `NaiveForecaster(strategy="mean", sp=12)` produced:

```python
model = craft("NaiveForecaster(strategy="mean", sp=12)")
```

→ `SyntaxError: invalid syntax. Perhaps you forgot a comma?`

This hits the common case: any estimator with a string hyperparameter (`strategy=`, `method=`, `averaging_method=`, `init_algorithm=`, …). Quote-free specs exported fine, confirming the trigger is precisely the embedded `"`.

## Fix

Use `json.dumps(value)` — it escapes embedded quotes/backslashes and its output is also a valid Python string literal, while keeping the existing double-quote style. Dict keys are now routed through `_format_value` too, so they get the same escaping (and non-string keys are no longer silently stringified).

## Testing

- New unit tests: embedded double quotes, backslashes, dict keys with quotes — each round-trips through `eval`
- New end-to-end test: instantiate `NaiveForecaster(strategy="mean", sp=12)`, export, `compile()` + `exec()` the generated code, and assert the recreated model's params
- `pytest tests/test_codegen.py` — 29 passed

### DIFF
--- a/src/sktime_mcp/tools/codegen.py
+++ b/src/sktime_mcp/tools/codegen.py
@@ -4,6 +4,7 @@ Code generation tool for sktime MCP.
 Generates Python code to recreate estimators and pipelines.
 """
 
+import json
 import keyword
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


Shashankss1205:
Merged into main as c6b2fd7 (local squash-merge; the API merge endpoint is blocked by the org's OAuth app restrictions). Closing.

PR #519, opened 2026-08-06
fix: reject empty data_handles in plot_series

Shashankss1205:
## Problem

`plot_series(data_handles=[])` raised a raw `IndexError` — `"list index out of range"` — leaked from sktime's plotting internals instead of a domain error (BUG-22 from the 2026-07-26 test sweep). `plotting.py` never checked for an empty list.

## Fix

Two layers:

- **Server-side**: `plot_series_tool` returns `{"success": false, "error": "data_handles must contain at least one data handle ID."}` before touching the executor.
- **Schema-side**: `"minItems": 1` on the `data_handles` property, so conformant MCP clients reject the call before it ever reaches the server.

## Testing

- New test: `plot_series_tool(data_handles=[])` returns a structured error
- `pytest tests/test_plotting.py` — 18 passed

### DIFF
--- a/src/sktime_mcp/server.py
+++ b/src/sktime_mcp/server.py
@@ -760,6 +760,7 @@ async def list_tools() -> list[Tool]:
                     "data_handles": {
                         "type": "array",
                         "items": {"type": "string"},
+                        "minItems": 1,
                         "description": "List of data handle IDs to plot (e.g., train, test, forecasts).",
                     },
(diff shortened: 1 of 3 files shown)


Shashankss1205:
Merged into main as ffe8510 (local squash-merge; the API merge endpoint is blocked by the org's OAuth app restrictions). Closing.

PR #520, opened 2026-08-06
fix: materialize generator results in call_method

Shashankss1205:
## Problem

`call_method` returned generators unmaterialized (BUG-24 from the 2026-07-26 test sweep). The result sanitization path had no generator branch, so generator results fell through to `repr()`:

```
call_method(handle_id=<SlidingWindowSplitter>, method_name="split", kwargs={"y_dataset": "airline"})
-> {"success": true, "result": "<generator object BaseSplitter.split at 0x726df47b6140>"}
```

Splitters are the headline use case in the tool's own description, yet `split`/`split_loc` were useless — `get_n_splits` was the only working path.

## Fix

Materialize generators to lists before sanitization:

```python
if inspect.isgenerator(result):
    result = list(result)
```

`sanitize_for_json` already handles lists, tuples, and ndarrays recursively, so the folds come back as JSON-safe nested int lists.

## Testing

- New `tests/test_call_method_generators.py`: `split` on a `SlidingWindowSplitter` over the airline dataset returns actual materialized folds (pairs of int index lists, train window length verified); non-generator results (`get_n_splits`) unchanged
- `pytest tests/test_call_method_generators.py` — 2 passed

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -636,6 +636,11 @@ class Executor:
 
             result = method(**kwargs)
 
+            # Materialize generators (e.g. splitter.split) so the caller gets
+            # the actual values instead of a useless repr string
+            if inspect.isgenerator(result):
+                result = list(result)
+
             from sktime_mcp.server import sanitize_for_json
 
             if hasattr(result, "to_dict"):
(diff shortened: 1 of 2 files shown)


Shashankss1205:
Merged into main as 685195e (local squash-merge; the API merge endpoint is blocked by the org's OAuth app restrictions). Closing.

PR #521, opened 2026-08-06
fix: surface dataset load errors in call_method kwargs

Shashankss1205:
## Problem

`call_method` silently swallowed an unresolvable `*_dataset` kwarg (BUG-25 from the 2026-07-26 test sweep). The kwarg-rewrite loop only acted `if data_res.get("success")` — on load failure it neither errored nor removed the key, so the raw string reached the method:

```
call_method(handle_id=<splitter>, method_name="get_n_splits",
            kwargs={"y_dataset": "no_such_dataset_zzz"})
-> "BaseWindowSplitter.get_n_splits() got an unexpected keyword argument 'y_dataset'"
```

The real cause (unknown dataset) was hidden behind a misleading signature error. The sibling `*_data_handle` branch already handled this correctly (`Unknown data handle: ...`).

## Fix

Mirror the `*_data_handle` branch: when the dataset fails to load, return the load error immediately, including the `available` dataset list when present:

```json
{"success": false, "error": "Unknown dataset: no_such_dataset_zzz", "available": [...]}
```

## Testing

- New `tests/test_call_method_dataset_kwargs.py`: unknown dataset returns the load error (not the signature error) with the available list; known dataset still resolves and the method executes
- `pytest tests/test_call_method_dataset_kwargs.py` — 2 passed

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -621,11 +621,18 @@ class Executor:
             for k, v in list(kwargs.items()):
                 if k.endswith("_dataset") and isinstance(v, str):
                     data_res = self.load_dataset(v)
-                    if data_res.get("success"):
-                        # Replace the kwarg with the actual data (e.g. y_dataset -> y)
-                        actual_key = k.replace("_dataset", "")
-                        kwargs[actual_key] = data_res["data"]
-                        del kwargs[k]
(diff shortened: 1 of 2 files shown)


Shashankss1205:
Merged into main as c20c9f7 (local squash-merge; the API merge endpoint is blocked by the org's OAuth app restrictions). Closing.

PR #524, opened 2026-08-08
fix: surface per-fold exceptions in evaluate instead of NaN metrics

Shashankss1205:
## Problem

`evaluate` masked per-fold exceptions and returned `success: true` with all-NaN metrics — the worst failure mode found in the 2026-07-26 test sweep (BUG-05, P0, silently-wrong answer).

`_run_evaluate` called sktime's `evaluate()` without `error_score`, so sktime's default `error_score=np.nan` swallowed every per-fold exception:

```json
{"success": true,
 "metrics": {"test_MeanAbsolutePercentageError": NaN},
 "fold_results": [{"test_MeanAbsolutePercentageError": NaN, "pred_time": NaN, "cutoff": null}, ...]}
```

Observed trigger: ThetaForecaster (which coerces to `PeriodIndex` internally) × a handle-loaded `DatetimeIndex` with `freq='MS'` — every fold raised `<MonthBegin> is not supported as period frequency`, and the response still said success. The same defect surfaces honestly in `predict` (`success: false` with the real error); `evaluate` was the only tool hiding it.

## Fix

Pass `error_score="raise"` so fold errors propagate to the existing exception handler and the tool returns `{"success": false, "error": "<real error>"}`. Applies to both the sync tool and the async job path (shared `_run_evaluate`).

## Testing

- New test reproduces the exact trigger (ThetaForecaster + MonthBegin-freq handle) and asserts the evaluation fails with a real error, and that no NaN metric can masquerade as a result
- `pytest tests/test_evaluate.py tests/test_evaluate_async.py tests/test_evaluate_summary.py` — 14 passed

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -117,7 +117,11 @@ def _run_evaluate(
         win = max(1, n - folds)
     cv = ExpandingWindowSplitter(initial_window=win, step_length=1, fh=[1])
 
-    results = evaluate(forecaster=instance, y=y, X=X, cv=cv, scoring=scoring)
+    # error_score="raise" — sktime's default (np.nan) swallows per-fold
+    # exceptions and reports success with all-NaN metrics
+    results = evaluate(
+        forecaster=instance, y=y, X=X, cv=cv, scoring=scoring, error_score="raise"
+    )
     if "estimator" in results.columns:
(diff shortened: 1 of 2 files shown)


Shashankss1205:
Merged into main as fa26f5f (local squash-merge; the API merge endpoint is blocked by the org's OAuth app restrictions). Closing.

PR #527, opened 2026-08-08
fix: validate cv_folds and initial_window instead of silently clamping

Shashankss1205:
## Problem

`evaluate` clamped `cv_folds` instead of validating it (BUG-16 from the 2026-07-26 sweep):

- `cv_folds=0` → `success: true`, 1 fold
- `cv_folds=-5` → `success: true`, 1 fold (response even echoes `"cv_folds_requested": -5`)
- `cv_folds=500` on n=144 → `success: true`, 143 folds — and the metric mean computed only over surviving folds, a quietly biased number

`initial_window >= n` was passed to the splitter unchecked.

## Suggested merge order

Merge after #524 — both touch `_run_evaluate` (different hunks; validation above, `error_score` at the call).

## Fix

- Tool layer: fast rejection of `cv_folds < 1` / `initial_window < 1` before any work (covers the async path's job creation too)
- `_run_evaluate`: validate `1 <= cv_folds <= n-1` and `1 <= initial_window < n` with the valid range and series length in the message; the silent clamp is gone

## Testing

- New tests: cv_folds 0 / -5 / 500 and initial_window=144 on airline all rejected with informative errors
- `pytest tests/test_evaluate*.py` — 17 passed

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -111,10 +111,20 @@ def _run_evaluate(
 
     n = len(y)
     if initial_window is not None:
+        if not 1 <= initial_window < n:
+            raise ValueError(
+                f"initial_window must be between 1 and n-1={n - 1} "
+                f"(series has {n} observations), got {initial_window}"
+            )
         win = initial_window
     else:
-        folds = max(1, min(int(cv_folds), max(1, n - 1)))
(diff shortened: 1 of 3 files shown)


Shashankss1205:
Merged into main as 91d46b1 (local squash-merge; the API merge endpoint is blocked by the org's OAuth app restrictions). Closing.

PR #528, opened 2026-08-08
fix: require data for update and roll back on failure

Shashankss1205:
## Problem

Two `update` defects from the 2026-07-26 sweep, combined here because they are the same function's two failure modes:

**BUG-17** — `update(estimator_handle=<fitted>)` with no data arguments returned `{"success": true, "message": "Estimator updated successfully"}`. Nothing happened.

**BUG-18** — sktime's `update()` mutates the live instance in place, so a REJECTED update destroyed the fitted estimator:

```
update(estimator_handle=..., y_dataset="airline")   -> success: false  (index-type mismatch)
predict(estimator_handle=...)
  -> "This instance of NaiveForecaster has not been fitted yet. Please call `fit` first."
```

…while the handle manager's `is_fitted` flag still read True — MCP state and sktime state disagreed.

## Fix

- `update` requires `y` (clear error naming `y_handle`/`y_dataset`)
- The instance is deep-copied before `update` and restored on exception via new `HandleManager.replace_instance`

## Testing

- New `tests/test_update_hardening.py`: no-data update rejected; failed update (DatetimeIndex-fitted × PeriodIndex airline) leaves the estimator predicting from its original state; successful update still works
- `pytest tests/` — full suite green

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -671,6 +671,21 @@ class Executor:
         if not self._handle_manager.is_fitted(handle_id):
             return {"success": False, "error": "Estimator not fitted"}
 
+        if y is None:
+            return {
+                "success": False,
+                "error": (
+                    "update requires new data — provide y_handle or y_dataset "
+                    "(and optionally X_handle/X_dataset)."
+                ),
+            }
+
+        # update mutates the live instance in place; snapshot fitted state so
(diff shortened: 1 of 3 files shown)


Shashankss1205:
Merged into main as f2790c3 (local squash-merge; the API merge endpoint is blocked by the org's OAuth app restrictions). Closing.

PR #542, opened 2026-08-08
fix: normalise loaded handles to PeriodIndex so seasonal forecasters work (#531)

### LINKED ISSUE #531: [BUG] Seasonal forecasting broken on user-loaded monthly data (MS DatetimeIndex vs PeriodIndex)
**Severity:** P0 — this is the single biggest UX blocker in the server today. Verified live over the MCP wire on 2026-08-09 (server at `main` `cbac09f`, sktime 1.0.1).

## Problem

Any user who loads their **own monthly data** and picks a **seasonal forecaster** (`sp > 1`) hits a dead end: `fit` succeeds, then `predict` fails.

```
load_data_source(monthly sales, 30 rows)     -> DatetimeIndex, freq "MS"   OK
fit(NaiveForecaster(strategy="last", sp=12), y_handle=<handle>)            OK  {"fitted": true}
predict(estimator_handle, horizon=6)
  -> {"success": false, "error": "<MonthBegin> is not supported as period frequency"}
```

Same failure with `AutoETS(auto=True, sp=12)`. The failure matrix (all confirmed live):

| | Demo dataset (PeriodIndex) | Loaded handle (DatetimeIndex "MS") |
|---|---|---|
| Non-seasonal (`sp=1`) | works | works (incl. intervals) |
| Seasonal (`sp>1`) | works | **fit OK, predict fails** |

## Root cause

`load_data_source` normalizes monthly data to a `DatetimeIndex` with `freq="MS"` (MonthBegin), while sktime's demo datasets carry a `PeriodIndex`. Seasonal forecasters that internally call `_pivot_sp` do `index.to_period(freq=MonthBegin)`, and pandas raises `<MonthBegin> is not supported as period frequency`. So `fit` (which doesn't pivot) succeeds and `predict` (which does) fails — a trap, because the user has no signal at fit time.

There is **no workaround inside the toolset** for datetime-indexed data; the only escapes are demo datasets, `sp=1` models, or integer-index loading.

## Related (same root cause)

- **NB-19** — `update` after a demo-dataset fit, with a `load_data_source` handle, fails the same way (demo PeriodIndex vs handle DatetimeIndex → object base Index deep in sktime).
- This is also the trigger class behind the former `evaluate` all-NaN masking bug (fixed in commit fa26f5f / #524, which now surfaces it *honestly* — but the underlying interop is still broken).

## Proposed fix

Normalize period-compatible `DatetimeIndex` handles to `PeriodIndex` at load time (in `load_data_source` / `format_data_handle`), so handle-loaded data behaves like demo data. One change resolves this issue, NB-19, and the BUG-05 trigger class. Alternatively, pre-validate at fit time and fail with a clear message instead of letting fit succeed.

## Re-verify

`load_data_source` 24+ monthly rows → `fit(NaiveForecaster(sp=12))` → `predict(horizon=6)` must return a forecast.

*(Catalogued as NB-20 / NB-19 in `MCP_TEST_FINDINGS.md`.)*

Shashankss1205:
Fixes #531.

## Problem

`load_data_source` produced a `DatetimeIndex` with freq `"MS"` (MonthBegin) for monthly data. Seasonal forecasters (`sp>1`) coerce the index to a `PeriodIndex` internally via `index.to_period(freq)` and raise `<MonthBegin> is not supported as period frequency` — so **fit succeeded and predict failed**, with no signal at fit time and no workaround inside the toolset. The same root cause broke `evaluate` on handle data and demo-fit + handle-update interop (NB-19).

## Fix

Normalise a regular `DatetimeIndex` to `PeriodIndex` at load, matching sktime's demo datasets (which carry a `PeriodIndex` and work end-to-end). Key detail: `to_period()` is called **with no argument** so pandas maps the offset to its period alias (`MS` → `"M"`); passing the inferred `"MS"` string back in re-raises the same error. No-op for irregular indexes, already-`PeriodIndex`, or non-datetime indexes.

Applied in `format_data_handle` (the auto-format-on-load path) and in the auto-format-disabled fallback, so it holds regardless of `SKTIME_MCP_AUTO_FORMAT`.

## Testing

- New `tests/test_periodindex_normalization.py` (7 tests): the helper across MS/inferred/irregular/period inputs; load yields a PeriodIndex; the **exact reported repro** (`fit(NaiveForecaster(sp=12), y_handle)` → `predict(horizon=6)`) now succeeds; seasonal `evaluate` on a handle returns real (non-NaN) metrics.
- Verified `split_data`, `save_data`, `plot_series`, `inspect_data` all still work on period-indexed handles (cutoff `2021-12`, 24/6 split, CSV 30 rows, 2-series plot).
- Full suite: **250 passed**.

## Verify over the wire (after server restart)

```
load_data_source(24+ monthly rows) -> fit(NaiveForecaster(sp=12), y_handle) -> predict(horizon=6)
```
must return a forecast (previously errored at predict).

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -49,6 +49,38 @@ def _get_demo_datasets() -> dict:
     return _DEMO_DATASETS
 
 
+def _to_period_index_if_possible(obj: Any) -> Any:
+    """Return *obj* with a ``PeriodIndex`` when its index is a regular datetime index.
+
+    Seasonal sktime forecasters coerce the series index to a ``PeriodIndex``
+    internally (``index.to_period(freq)``) and raise on offset frequencies such
+    as ``MonthBegin`` ("MS"), which is what ``load_data_source`` produces for
+    monthly data. Demo datasets carry a ``PeriodIndex`` and work, so we
+    normalise handle-loaded data to match. ``to_period()`` is called with no
+    argument so pandas maps the offset to its period alias (MS -> "M"); passing
+    the offset string back in would re-raise the same error.
+
+    No-op for non-datetime indexes, ``PeriodIndex`` already, or an index with no
+    determinable frequency.
+    """
+    if obj is None or not hasattr(obj, "index"):
+        return obj
+    idx = obj.index
+    if isinstance(idx, pd.PeriodIndex) or not isinstance(idx, pd.DatetimeIndex):
+        return obj
+    try:
+        if idx.freq is None:
+            inferred = pd.infer_freq(idx)
+            if inferred is None:
+                return obj
+            idx = pd.DatetimeIndex(idx, freq=inferred)
+        converted = obj.copy()
+        converted.index = idx.to_period()
+        return converted
+    except (ValueError, TypeError):
+        return obj
+
+
 def _get_index_frequency_metadata(
     index: pd.Index,
     fallback: str | None = None,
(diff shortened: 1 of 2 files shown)


Shashankss1205:
Closing: this PR's content is already on main as commit 185250d (identical patch-id; it was landed directly rather than through the PR merge button). Nothing further to merge.

PR #543, opened 2026-08-08
fix: report evicted handles as "evicted", not "not found" (#532)

### LINKED ISSUE #532: [BUG] Silent bulk handle eviction reports "not found" — user state vanishes with no warning
**Severity:** P1 — silent loss of user state. Verified live 2026-08-09 (server at `main` `cbac09f`).

## Problem

Both handle stores silently bulk-evict the oldest entries when they hit their cap, and the caller only learns about it when a later call returns a generic **"not found"** — indistinguishable from a typo.

Live repro: a data handle created at the start of a session was gone after ~52 subsequent `load_data_source` calls:

```
inspect_data(data_handle="data_bdaec9a2")
  -> {"success": false, "error": "Data handle 'data_bdaec9a2' not found"}
```

## Root cause

- **Estimator store** — `HandleManager._cleanup_oldest` ([`handles.py`](../blob/main/src/sktime_mcp/runtime/handles.py)) deletes the 10 oldest outright when `len >= max_handles` (default 100).
- **Data store** — `Executor._cleanup_oldest_data` ([`executor.py`](../blob/main/src/sktime_mcp/runtime/executor.py)) deletes the oldest ~20% at once, emitting only a `logger.debug` — nothing reaches the caller.

## Proposed fix

Distinguish *evicted* from *never existed*. Keep a tombstone set in both stores and return, e.g., `"Data handle 'X' was evicted (limit N reached); reload the source."` Better still: expose the limits + current usage via `list_handles` / `list_available_data`, and/or make eviction opt-in (raise at the cap). Never silently discard user state.

## Re-verify

Create handles past the cap, then use the first one — the error must say *evicted*, not *not found*.

*(Catalogued as BUG-08 in `MCP_TEST_FINDINGS.md`.)*

Shashankss1205:
Fixes #532.

## Problem

Both handle stores silently bulk-evict the oldest entries at their cap (estimator store: 10 oldest via `HandleManager._cleanup_oldest`; data store: oldest ~20% via `Executor._cleanup_oldest_data`), emitting only a `logger.debug`. The caller then got a generic **"not found"** for a handle it had legitimately created — indistinguishable from a typo. (Confirmed live: a session-start data handle vanished after ~52 loads with a plain not-found.)

## Fix

Track evicted ids in a bounded tombstone `deque(maxlen=1024)` in both stores, and route every not-found site through a shared helper:

- `HandleManager.describe_missing(id)` — used by `get_instance`/`get_info` (raise), `evaluate`, `export_code`, `save_model`, `release_handle`.
- `Executor.data_handle_missing(id)` — returns the error string + capped available-handles summary, used by `inspect_data`, `split_data`, `save_data`, `transform_data`, `plot_series`, and the executor's own `format_data_handle` / `release_data_handle`.

When the id was evicted, the message becomes *"'<id>' was evicted (handle limit N reached); reload the source / re-create it with instantiate"* instead of "not found". Eviction is now logged at info level.

## Testing

- New `tests/test_eviction_tombstones.py` (4 tests): estimator + data eviction messages, `get_instance` raises the eviction message, and `inspect_data` surfaces it end-to-end; a never-seen id still reads as a plain "not found".
- Full suite: **254 passed**.

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -9,6 +9,7 @@ import asyncio
 import inspect
 import logging
 import uuid
+from collections import deque
 from typing import Any
 
 import pandas as pd
@@ -196,6 +197,8 @@ class Executor:
         self._handle_manager = get_handle_manager()
         self._job_manager = get_job_manager()
         self._data_handles: dict[str, Any] = {}
+        # Tombstones for data handles evicted under the cap (see _cleanup_oldest_data).
+        self._evicted_data: deque[str] = deque(maxlen=1024)
         from sktime_mcp.config import settings
 
         self._max_data_handles = settings.max_data_handles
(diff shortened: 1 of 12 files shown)


Shashankss1205:
Closing: this PR's content is already on main as commit 72bfd4a (identical patch-id; it was landed directly rather than through the PR merge button). Nothing further to merge.

PR #544, opened 2026-08-08
fix: close load_data_source validation gaps (#533)

### LINKED ISSUE #533: [BUG] load_data_source validation gaps: non-numeric target passes, 1–2 row load blocked, file loader leaks internals
**Severity:** P1 (dtype) + P2 (row-count, error quality). Verified live 2026-08-09.

`load_data_source` validation runs but doesn't check the things `fit` actually needs, and several failure paths leak pandas internals. Three related findings:

## NB-07 — non-numeric (object-dtype) target passes validation silently (P0-ish)

```
load_data_source(config={"type":"pandas",
  "data":{"date":["2024-01-01","2024-02-01","2024-03-01"], "value":["1","2","3"]},
  "time_column":"date","target_column":"value"})
-> {"success": true, "dtypes": {"value": "object"}, "validation": {"valid": true}}
```

`inspect_data` then returns categorical stats (`{count, unique, top, freq}`, no mean/min/max). It fails much later, inside `fit`. Same gap for object-dtype exogenous columns.
**Fix:** error, or at least warn, on a non-numeric target/exog dtype at load.

## NB-06 — 1–2 row series cannot be loaded at all (P1)

```
load_data_source(... 1 row ...)
-> {"success": false, "error": "Need at least 3 dates to infer frequency", "error_type": "ValueError"}
```

This is a leaked pandas internal, and it's inconsistent: `split_data` happily *creates* 1-row handles, so 1-row data is representable — only loading is blocked.
**Fix:** load with `freq=None` + a warning, or a clear "at least 3 timestamped rows required to infer frequency" message.

## NB-11 — file loader leaks `parse_dates` internals for a bad time_column (P2)

```
load_data_source(config={"type":"file","path":"...csv","time_column":"index", ...})
-> {"error": "Error reading CSV file: Missing column provided to 'parse_dates': 'index'"}
```

The pandas-source path gives a clean *"Time column 'x' not found in data"*; the file path should match it (and ideally list available columns).

*(Catalogued as NB-07 / NB-06 / NB-11 in `MCP_TEST_FINDINGS.md`. Related to existing #313.)*

Shashankss1205:
Fixes #533.

Three related `load_data_source` validation gaps, all verified in the sweeps:

## NB-07 — non-numeric target passed validation silently
An object-dtype target (`value: ["1","2","3"]`) returned `valid: true` and failed much later inside `fit`. `validate()` now **warns** when the target column is non-numeric, naming the column and dtype. The check is forwarded to file sources too — `file_adapter.validate` now passes the target/exog config into the reused `PandasAdapter` validation (previously it constructed one with `{"data": data}` only, so the target was unknown).

## NB-06 — 1–2 row series could not load
`pd.infer_freq` requires ≥3 points and raises `"Need at least 3 dates to infer frequency"`; three call sites hit it unguarded. Added a `_safe_infer_freq` helper that returns `None` for short series, so a tiny (but valid) series loads with no frequency instead of crashing.

## NB-11 — file loader leaked pandas parse_dates internals
A bad `time_column` produced `"Missing column provided to 'parse_dates': 'index'"`. Now the time column's existence is checked up front with a clean *"Time column 'x' not found in data. Available columns: [...]"* message, and `parse_dates=[col]` is no longer wired into `read_csv` — the column is parsed to datetime in `load()` after the existence check (matching the pandas-source path).

## Testing

- New `tests/test_load_validation.py` (6 tests): object-dtype target warns / numeric doesn't; 1- and 2-row series load; bad file `time_column` gives a clean not-found without `parse_dates` in the message and lists available columns; bad `target_column` lists available.
- Full suite: **260 passed**.

### DIFF
--- a/src/sktime_mcp/data/adapters/file_adapter.py
+++ b/src/sktime_mcp/data/adapters/file_adapter.py
@@ -78,6 +78,11 @@ class FileAdapter(DataSourceAdapter):
 
         # Set time index
         time_col = self.config.get("time_column")
+        if time_col is not None and time_col not in df.columns:
+            available = ", ".join(repr(c) for c in df.columns)
+            raise ValueError(
+                f"Time column {time_col!r} not found in data. Available columns: [{available}]"
+            )
         if time_col and time_col in df.columns:
             if self.config.get("parse_dates", True):
                 with contextlib.suppress(Exception):
@@ -108,7 +113,9 @@ class FileAdapter(DataSourceAdapter):
 
         # Determine frequency for metadata
         if isinstance(df.index, pd.DatetimeIndex):
-            freq_str = str(df.index.freq) if df.index.freq else pd.infer_freq(df.index)
+            from .pandas_adapter import _safe_infer_freq
+
+            freq_str = str(df.index.freq) if df.index.freq else _safe_infer_freq(df.index)
         else:
             freq_str = "Integer"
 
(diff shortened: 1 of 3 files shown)


Shashankss1205:
Closing: this PR's content is already on main as commit 052a4c5 (identical patch-id; it was landed directly rather than through the PR merge button). Nothing further to merge.

PR #526, opened 2026-08-08
fix: expand and resolve local paths in save_model/load_model

Shashankss1205:
## Problem

Found in the 2026-08-08 sweep:

1. `save_model(path="~/model")` created a **literal `~` directory** in the server cwd (the user's repo checkout) — an `rm -rf ~` cleanup footgun — while reporting success.
2. Relative paths silently resolved against the server's cwd, writing MLflow artifacts into the repo working tree, and `saved_path` echoed the unresolved input so the caller could not learn the real location.

## Fix

New `resolve_model_path`: `os.path.abspath(os.path.expanduser(path))` for local paths; MLflow URIs (`runs:/`, `models:/`, `mlflow-artifacts:/`, any `scheme://`) pass through untouched. `saved_path` and the success message now echo the resolved absolute path. `load_model` applies the same resolution so round-trips agree.

## Testing

- New tests: tilde expansion, relative→absolute, MLflow URI passthrough
- `pytest tests/test_save_model_fitted_check.py` — 6 passed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #545, opened 2026-08-08
fix: refuse to silently overwrite existing files in save_data (#539)

### LINKED ISSUE #539: [BUG] save_data silently overwrites existing files
**Severity:** P2. Verified live 2026-08-08.

## Problem

`save_data` silently overwrites an existing file with no indication.

```
save_data(data_handle=<20-row handle>, path="out.csv", format="csv")  -> {"success": true, "rows": 20}
save_data(data_handle=<3-row handle>,  path="out.csv", format="csv")  -> {"success": true, "rows": 3}
```

The second call replaces the file with no `overwritten` flag or warning. This is inconsistent with `save_model`, which refuses to write into a non-empty path.

## Proposed fix

Return `overwritten: true` (or warn), or require an explicit `overwrite=true` opt-in — matching `save_model`'s refuse-on-non-empty behavior.

*(Catalogued as NB-12 in `MCP_TEST_FINDINGS.md`.)*

Shashankss1205:
Fixes #539.

## Problem

`save_data` silently overwrote an existing file with no indication — inconsistent with `save_model`, which refuses to write into a non-empty path.

## Fix

- New `overwrite: bool = False` parameter (plumbed through the MCP schema and dispatch). An existing target is refused with *"File already exists: '<path>'. Pass overwrite=true to replace it..."*; with `overwrite=true` the write proceeds and the response reports `overwritten: true`. Fresh writes report `overwritten: false`.
- Also `expanduser()` the path, so `save_data(path="~/x")` no longer creates a literal `~` directory in the server cwd — the same footgun fixed for `save_model` in #526.

## Testing

- New `test_save_refuses_existing_file`: first write succeeds, second to the same path is refused, third with `overwrite=true` succeeds and reports `overwritten: true`.
- Updated the existing csv/json tests to fresh paths (they previously wrote into a pre-created `NamedTemporaryFile`, which the new guard correctly refuses) and assert `overwritten: false`.
- Schema lint (`test_schema_types`) passes with the new typed `overwrite` property.
- Full suite: **261 passed**.

### DIFF
--- a/src/sktime_mcp/server.py
+++ b/src/sktime_mcp/server.py
@@ -751,6 +751,14 @@ async def list_tools() -> list[Tool]:
                         "enum": ["csv", "parquet", "json"],
                         "default": "csv",
                     },
+                    "overwrite": {
+                        "type": "boolean",
+                        "description": (
+                            "Replace the file if it already exists. Default false — an "
+                            "existing file is not clobbered unless this is true."
+                        ),
+                        "default": False,
+                    },
                 },
                 "required": ["data_handle", "path"],
             },
@@ -1116,6 +1124,7 @@ async def call_tool(name: str, arguments: dict[str, Any]) -> list[TextContent]:
                 data_handle=arguments["data_handle"],
                 path=arguments["path"],
                 format=arguments.get("format", "csv"),
+                overwrite=arguments.get("overwrite", False),
             )
 
         elif name == "auto_format_on_load":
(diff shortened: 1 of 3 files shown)


Shashankss1205:
Closing: this PR's content is already on main as commit b2a63d5 (identical patch-id; it was landed directly rather than through the PR merge button). Nothing further to merge.

Issue #540, opened 2026-08-08
[Security] Trust boundary: instantiate eval, non-estimator handles, call_method dunders, run_command host + unbounded output


Shashankss1205:
**Severity:** P1 (security / trust boundary). Verified live from the 2026-07-26 sweep.

## Context

The server already exposes `run_command` (arbitrary bash by design), so the caller is nominally fully trusted. **The finding is that three other tools grant equivalent power while advertising narrow scope** — any policy layer that allowlists `instantiate` or `call_method` while blocking `run_command` is bypassable. This is an umbrella issue: **decide the intended trust posture first**, then the per-tool fixes below follow.

## BUG-09 — `instantiate` is an unsandboxed `eval`

`instantiate` → `craft(spec)` → `eval(spec, ...)`. All returned `success: true`:
```
instantiate(spec="__import__('os').system('echo pwned')")   -> handle, estimator: "int"
instantiate(spec="open('/etc/hostname').read()")            -> handle, estimator: "str"
```
Tool description: *"Create an estimator or pipeline instance."*
**Fix:** document it as arbitrary code execution (treat as `run_command` in any policy), **or** validate the spec with an AST allowlist (Call/Name/Constant/BinOp over registry names) before `craft`.

## BUG-10 — no post-craft type check; non-estimators get `est_` handles

```
instantiate(spec="42")            -> est_..., estimator: "int"
instantiate(spec="[1,2,3]")       -> est_..., estimator: "list"
```
These fail confusingly downstream (see the evaluate scitype issue).
**Fix:** after `craft`, assert the result is an sktime `BaseObject` (or at least has `get_params`); reject otherwise with *"spec did not produce an sktime object, got int"*. (Related: #219.)

## BUG-11 — `call_method` applies zero method-name filtering

Only checks `hasattr`, then `getattr` + call. No denylist/allowlist/dunder filter:
```
call_method(handle_id=<NaiveForecaster>, method_name="__reduce__", kwargs={})
  -> success, dumps the full internal __dict__
```
On a **fitted** estimator this dumps `_y` and `_X` — the training data — to any caller. `__class__`, `_private` methods, etc. are all reachable.
**Fix:** denylist `method_name.startswith("_")` except an explicit allowlist of useful dunders, or allowlist per `object_type`.

## BUG-12 — `run_command` runs on the host, description falsely claims a container

[...]

PR #551, opened 2026-08-08
fix: conservative trust-boundary hardening (#540)

Shashankss1205:
Addresses #540 — the **conservative, non-controversial** parts of the trust-boundary umbrella.

## Scope note

**BUG-09 (sandboxing `instantiate`/`craft`) is intentionally left out.** The server already exposes `run_command` (arbitrary host execution by design), so the caller is nominally trusted; locking down `craft` specs with an AST allowlist risks breaking legitimate composites and is a posture decision for the maintainers. This PR does the parts that are pure improvements regardless of posture.

## BUG-11 — call_method exposed private/dunder methods
`call_method(method_name="__reduce__")` returned the full `__dict__`, dumping the fitted `_y`/`_X` **training data** to any caller. `call_method` now rejects any `method_name` starting with `_`, except a small allowlist (`__call__` for callable metrics/aligners, `__len__`, `__repr__`, `__str__`). Public methods are unaffected.

## BUG-10 — instantiate accepted non-estimators
`instantiate("42")` → an `est_` handle of type `int` that failed confusingly downstream. After `craft`, the result must be an sktime object (skbase `BaseObject`, or duck-typed `get_params` + `get_class_tag`); otherwise a clear *"Spec did not produce an sktime estimator, got int"*.

## BUG-12 — run_command description was false
It claimed the command runs "inside the sktime container". Corrected to say it runs on the **host**, in the server's working directory, as the server user.

## BUG-23 — run_command output was unbounded
`seq 1 100000` returned ~689k chars. Output is capped at ~20k chars (head+tail) with a `truncated` flag, and a non-zero exit now includes an `"error"` key like every other tool (also closes N-24).

## Testing

- New `tests/test_security_hardening.py` (14 tests): `__reduce__`/`__class__`/private methods blocked, public methods work; `42`/`[1,2,3]`/`None` rejected, real estimator still instantiates; large output truncated, small output intact, non-zero exit has `error`.
- Full suite: **303 passed**.

### DIFF
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -87,6 +87,29 @@ def _to_period_index_if_possible(obj: Any) -> Any:
 # flood the client with ~30KB+ of inline JSON.
 _MAX_PREDICTION_ROWS = 500
 
+# Dunder methods that are safe and useful to call via call_method (e.g. __call__
+# for callable metrics/aligners). Everything else starting with "_" is blocked
+# (BUG-11) — notably __reduce__/__class__/__getattribute__ and private methods.
+_ALLOWED_DUNDERS = frozenset({"__call__", "__len__", "__repr__", "__str__"})
+
+
+def _is_sktime_object(obj: Any) -> bool:
+    """True if *obj* is a genuine sktime estimator/object, not a bare value.
+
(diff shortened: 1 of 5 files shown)


Shashankss1205:
Closing: this PR's content is already on main as commit 333ff33 (identical patch-id; it was landed directly rather than through the PR merge button). Nothing further to merge.

Issue #471, opened 2026-06-03
Phase 2: Instantiation & Composition Tools


Shashankss1205:
The goal of this phase is to cleanly isolate and refactor the **Instantiation & Composition Tools**. These tools manage the creation, configuration, structural validation, and lifecycle of estimators and pipelines in the server's memory. This phase will ensure type safety, input parameter validation, and resource cleanup for model handles used by the client.

### 1. `instantiate`
Creates an executable estimator or pipeline instance. If creating a pipeline, it validates the structural compatibility and type safety of the chained components prior to instantiation. Registers the created object in the server memory and returns a unique handle.

**Arguments:**
* `estimator` *(str, optional)*: Name of a single estimator class to instantiate (e.g., `"ARIMA"`).
* `params` *(dict, optional)*: Hyperparameters for the single estimator.
* `components` *(list[str], optional)*: List of estimator class names in pipeline order for composition (e.g., `["Imputer", "ARIMA"]`).
* `params_list` *(list[dict], optional)*: Optional list of parameter dictionaries for each component in the pipeline.

*Note: Exactly one of `estimator` or `components` must be provided.*

**Returns:**
* `dict`: `{"handle": str, "estimator": str, "params": dict | list[dict], "warnings": list[str]}`

### 2. `list_handles`
Lists all active estimator and pipeline handles currently stored in the server's memory.

**Arguments:**
* *None*

**Returns:**
* `dict`: `{"handles": list[dict], "count": int}`

### 3. `release_handle`
Releases a registered model handle and frees its associated resources from memory.

**Arguments:**
* `handle` *(str, required)*: The unique handle ID to release.

**Returns:**
* `dict`: `{"success": bool, "handle": str, "message": str}`

## Implementation Notes
* **`instantiate`**: 
  - For single estimators: Locate the class via the registry, validate parameters, construct the object, and register it.
  - For pipelines: Check pipeline order and compatibility of components (e.g., ensuring transformers precede forecasters/classifiers). If valid, construct the pipeline object, register it with the `HandleManager`, and return the handle. If validation fails, return an informative error listing the structural conflicts.
* **`list_handles`**: Query the `HandleManager` for active handles, returning details such as estimator class name, hyperparameters, and fitted status.
* **`release_handle`**: Call `release_handle` on the `HandleManager` to delete the handle and make its memory available for garbage collection.

## Acceptance Criteria
- [ ] Implement the `instantiate`, `list_handles`, and `release_handle` tools as defined above.
[...]

PR #555, opened 2026-09-01
[BUG] lock `HandleManager` against concurrent tool calls

### LINKED ISSUE #554: [BUG] `HandleManager` has no locking — `KeyError`/`RuntimeError` under concurrent tool calls
`HandleManager` (`runtime/handles.py`) mutates `self._handles` with no synchronisation, while `JobManager` (`runtime/jobs.py:145`) holds a `threading.Lock()`.

The handle store is reachable from worker threads: `fit_async` runs `self.fit` in a `ThreadPoolExecutor` (`executor.py:988`) and `fit` writes via `mark_fitted` (`executor.py:507`); `update` writes via `replace_instance` (L897) — while the asyncio thread keeps serving other tool calls.

Two concrete races:

- `_cleanup_oldest` (L125) snapshots `sorted(self._handles.items())`, then `del`s in a loop. Two `create_handle` calls at the cap select the same victims and the second delete raises an unhandled `KeyError`.
- `list_handles` (L117) iterates `self._handles.values()`; an eviction mid-iteration gives `RuntimeError: dictionary changed size during iteration`.

`mark_fitted`, `is_fitted`, `replace_instance` and `release_handle` are all check-then-act.

Repro with no instrumentation — 16 threads, cap 30, `sys.setswitchinterval(1e-6)`:

```
trials with error: 20/20
  RuntimeError: 9728
  KeyError: 444
```

At the default 5 ms switch interval on an idle machine it stays hidden, which is probably why it hasn't surfaced yet.

`Executor._data_handles` (`executor.py:253-278`) has the same list-then-delete and check-then-act shape. Correcting myself after a closer look: I can't find a live race there, so it probably doesn't need the same fix. Nothing off the event loop reaches it — the `run_in_executor` callables (`self.predict`, `_run_evaluate`, `run_fit`, `adapter.load`) all take pre-resolved data and never read `_data_handles` — and none of its critical sections span an `await`. `HandleManager` differs because `fit` runs inside the pool and writes through `mark_fitted`.

This is the "thread-safe lifecycle management of all estimator/pipeline handles" criterion in #471.

PR to follow.

OfficialAbhinavSingh:
#### Reference Issues/PRs

Fixes #554.

`HandleManager` mutated `self._handles` with no synchronisation, while `JobManager` (`runtime/jobs.py:145`) already holds a `threading.Lock()`. The store is reached from worker threads — `fit_async` runs `fit` in a `ThreadPoolExecutor` (`executor.py:988`) and `fit` writes via `mark_fitted` (`executor.py:507`) — while the asyncio thread keeps serving other tool calls.

Two races followed:

- `_cleanup_oldest` snapshotted `sorted(self._handles.items())` and then deleted in a loop, so two `create_handle` calls at the cap picked the same victims and the second delete raised `KeyError`.
- `list_handles` iterated `self._handles.values()`, so an eviction mid-walk raised `RuntimeError: dictionary changed size during iteration`.

Guards the accessors with an `RLock` — reentrant because `create_handle` evicts while already holding it. No behaviour change single-threaded.

No, `threading` is stdlib.

- Whether locking `describe_missing` is wanted, given `get_instance`/`get_info` call it while already holding the lock. Safe under `RLock`, but it reads oddly.
- `Executor._data_handles` (`executor.py:253-278`) has the same list-then-delete and check-then-act shape, but I went looking and could not find a live race there, so I've deliberately not touched it. Nothing off the event loop reaches it: the `run_in_executor` callables (`self.predict`, `_run_evaluate`, `run_fit`, `adapter.load`) all take pre-resolved data and never read `_data_handles`, and none of its critical sections span an `await`. `HandleManager` differs because `fit` runs *inside* the pool and writes through `mark_fitted`. Happy to be told I've missed a path.

#### Any other comments?

The three tests fail on `main` and pass with the change; I checked both directions and reran them 20x for flakiness. The churn test sets `sys.setswitchinterval(1e-6)` and restores it — at the 5 ms default the race stays hidden on an idle machine, which is probably why it hasn't surfaced.

Full suite green locally: **312 passed in 31s**. Worth knowing `make check` is already red on `main` for unrelated files — 7 lint errors in `test_load_validation`, `test_save_model_fitted_check`, `test_transform_deadends`. Happy to clear those separately.

Unrelated dev-env note in case it's useful: `numba` isn't in the `dev` extra, and without it `test_classifier_example_runs` falls back to pure-Python DTW and effectively never finishes — it took me a while to work out that wasn't my change. Installing `numba` takes the suite from "hangs" to 31s.

I used Claude Code to help investigate and draft this; the repro and test results are my own verification.

#### PR checklist

### DIFF
--- a/src/sktime_mcp/runtime/handles.py
+++ b/src/sktime_mcp/runtime/handles.py
@@ -5,6 +5,7 @@ Manages references to instantiated estimator objects.
 """
 
 import logging
+import threading
 import uuid
 from collections import deque
 from dataclasses import dataclass, field
@@ -46,18 +47,24 @@ class HandleManager:
         # Tombstones: ids evicted to stay under the cap, so a later lookup can
         # say "evicted" instead of an indistinguishable "not found".
         self._evicted: deque[str] = deque(maxlen=1024)
+        # Handles are read and written from worker threads (fit_async runs fit
+        # in a thread pool) while the asyncio thread serves other tool calls.
+        # Reentrant because create_handle evicts while already holding it.
+        self._lock = threading.RLock()
 
     def describe_missing(self, handle_id: str) -> str:
         """Message for a handle that isn't present — distinguishes evicted from unknown."""
-        if handle_id in self._evicted:
-            return (
-                f"Estimator handle '{handle_id}' was evicted (handle limit "
-                f"{self._max_handles} reached); re-create it with instantiate."
-            )
-        return f"Handle not found: {handle_id}"
+        with self._lock:
+            if handle_id in self._evicted:
+                return (
+                    f"Estimator handle '{handle_id}' was evicted (handle limit "
+                    f"{self._max_handles} reached); re-create it with instantiate."
+                )
+            return f"Handle not found: {handle_id}"
 
     def was_evicted(self, handle_id: str) -> bool:
-        return handle_id in self._evicted
+        with self._lock:
+            return handle_id in self._evicted
 
     def create_handle(
         self,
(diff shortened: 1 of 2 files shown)


Shashankss1205:
Verified: the three new tests fail on current main with KeyError and pass with this change; full suite green; ruff clean. The RLock scope (HandleManager only) is right, because Executor._data_handles is only touched on the event-loop thread. Merging. Closes #554 and completes the last item of #471.

Shashankss1205:
Landed on main as def49b7 (squash-merged locally because the org policy rejects the merge API for this token). Thanks!

PR #559, opened 2026-09-08
fix: ruff PTH errors, format drift, bump pre-commit ruff to 0.16.6

### LINKED ISSUE #558: CI-CD checks failing
ruff check + format failing on main, blocking ci

CI lint stage is red. `ruff check .` fails with 7 PTH errors in the test files (os.path.join/open/os.getcwd instead of pathlib), and `ruff format --check .` wants to reformat 9 files.

samudraneel05:
Aligns pre-commit ruff version with what CI installs.

#### Reference Issues/PRs

Fixes #558 

ci lint is red on main. two fixes:
ruff check . gives 7 PTH errors in test files, which are fixed by converting os.path.join/open -> pathlib
ruff format run -> 9 files reformatted
bump pre-commit ruff 0.3.4 -> 0.16.6 so local and ci agree on rules
`ruff check .` and `ruff format --check .` both pass.

No

Just worth to see if checks are green, no material code changes

##### For all contributions
- [x] I've updated the existing example scripts or provided a new one to showcase how my tool works in `examples/`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.3.4
+    rev: v0.16.6
     hooks:
       - id: ruff
         args: [ --fix ]
--- a/src/sktime_mcp/runtime/executor.py
+++ b/src/sktime_mcp/runtime/executor.py
@@ -213,9 +213,7 @@ def _run_evaluate(
 
     # error_score="raise" — sktime's default (np.nan) swallows per-fold
     # exceptions and reports success with all-NaN metrics
-    results = evaluate(
-        forecaster=instance, y=y, X=X, cv=cv, scoring=scoring, error_score="raise"
-    )
+    results = evaluate(forecaster=instance, y=y, X=X, cv=cv, scoring=scoring, error_score="raise")
     if "estimator" in results.columns:
         results = results.drop(columns=["estimator"])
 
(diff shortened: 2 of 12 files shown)


Shriprasad-P (approved):
Reviewed the Ruff 0.16.6 alignment and pathlib conversions. `ruff check .`, `ruff format --check .`, and both Ruff pre-commit hooks pass with 0.16.6; the nine changed test modules pass (43 tests). No substantive findings.

Shriprasad-P (approved):
Reviewed the Ruff/pre-commit alignment and the related pathlib conversions. The update brings the local pre-commit Ruff version in line with the version currently resolved in CI, and I did not find any behavioral regressions in the changed tests.

I also checked the functional path conversions and the affected test modules passed locally. No blocking issues from my side.

Shriprasad-P (commented):
## Review

Mostly lint/format drift from bumping `ruff` pre-commit `v0.3.4` → `v0.16.6`, plus PTH-style path cleanups.

Looks mechanical and low-risk. Worth confirming CI green on the new ruff version across the whole tree (new rules can surprise). Fine to land as hygiene.

Me-Priyank:
Reproduced this on a clean checkout of `f6fb797`: `ruff check .` gives 7 errors and `ruff format --check .` wants 8 files; with the patch both pass, under ruff 0.16.6 and 0.16.8. I also diffed the AST of `executor.py` and `evaluate.py` before and after, identical, so those two are formatting-only.

One thing I noticed while digging: pre-commit pins `v0.3.4` but `pyproject.toml:48` has `ruff>=0.1.0`, so CI installs whatever's newest. Under the old pin main shows 14 extra `UP038` errors, that rule was removed in ruff 0.13.0. Might be worth bounding the dev extra here too, otherwise the drift comes back.

